PR #10, opened 2018-11-01
WIP Fix/json schema type annotations string and array

AlexanderMann:
# WIP


target_postgres/postgres.py
```diff
@@ -9,6 +9,8 @@
 import arrow
 from psycopg2 import sql
 
+import target_postgres.json_schema as JSONSchema
```
awm33: I actually think keeping it as `json_schema` would be more pythonic. Capitalized words are generally used for classes, which below made me thing you had created a JSONSchema class.

target_postgres/postgres.py
```diff
@@ -9,6 +9,8 @@
...
+
```
awm33: This is one thing it pep 8 I kind of like but generally there are three import groups with a line in between:
```
built in packages/modules

3rd-party packages/modules

package internal packages/modules
```

```
import io
import re
import csv
import uuid
import json
from functools import partial
from copy import deepcopy

import arrow
from psycopg2 import sql

from target_postgres import json_schema
from target_postgres.singer_stream import (
    SINGER_RECEIVED_AT,
    SINGER_BATCHED_AT,
    SINGER_SEQUENCE,
    SINGER_TABLE_VERSION,
    SINGER_PK,
    SINGER_SOURCE_PK_PREFIX,
    SINGER_LEVEL
)
```

awm33:
@AlexanderMann Given that there is an fairly immediate goal of sharing some of this code with redshift, and potentially other SQL-based targets (mysql, snowflake), the idea going forward is to be able to override functions on the PostgresTarget class (and probably rename it to a more generic SQL target class). See the `redshift` branch, which we may go in a different direction, but you can see how the differences are handled.

I like the idea of pulling out some of the JSON Schema functionality in to a separate module. Some functions may be able to be moved over to the new json_schema module, but things like `json_schema_to_sql` would change for other databases types, and should stay on the target object so that inheritance can change that behavior.

I think some of the other JSON Schema issues (like adding $ref and anyOf) may benefit from having this separate module. If something related to JSON Schema seems as though it would be the same across all SQL-like targets, or those differences could be clear arguments to the function, then I think it's a candidate for moving over to the new module.

target_postgres/postgres.py
AlexanderMann: Yeah, the reason I _didn't_ do that here was that the variable `json_schema` is...everywhere...in this file. Due to this, I figured it'd be far simpler to have a separate PR which only makes the correction to the names as a pure refactor, thus keeping the meat of this PR tiny and focused. I'm happy to make that change here as well in a single comment!

target_postgres/postgres.py
AlexanderMann: Easily done!

target_postgres/postgres.py
awm33: Hmm true. I also use the variable `json_schema` a lot in taps. We could rename the module to something like `json_schema_utils` or import the functions directly.

awm33:
@AlexanderMann One small thing to add. We get to remove "Types must be arrays" from Limitations in the README :)

target_postgres/postgres.py
```diff
@@ -736,7 +743,7 @@ def add_column(self, cur, table_schema, table_name, column_name, data_type, defa
     def merge_put_schemas(self, cur, table_schema, table_name, existing_schema, new_schema):
         new_properties = new_schema['properties']
         existing_properties = existing_schema['properties']
-        for prop, json_schema in new_properties.items():
+        for prop, _ in new_properties.items():
```
awm33: We can actually change this to:
`for prop in new_properties:`
(I realize this was like this before, but thought I'd mention it plus I know you are diving deep into python)

Issue #13, opened 2018-11-01
JSON Schema $ref is not supported


awm33:
- Add support for self referencing `$ref` (reference is within the same document).
- Identify level of effort for resolving HTTP(S) `$ref`s. Then comment and collaborate to determine if level of effort meets benefit.

AlexanderMann:
@awm33 I think we can probably close this as we now have basic `$ref` support. When we have a feature request/run into more advanced `$ref`s usage, I think we can revisit our approach at that time.

awm33:
Agreed

Issue #7, opened 2018-10-29
Test more data cases


awm33:
Test out more variations of data. Look to real world sources for examples.

awm33:
A perfect example of why more data cases are needed:

https://github.com/datamill-co/target-postgres/issues/2

I always lowercase and snakecase my data field names, this broke on camel case on redshift. Redshift lowercases the fieldnames, postgres does not. This specific example only effects Redshift, but you can see we need more cases.

AlexanderMann:
Example case that has not been tested:

```
{'type': 'any-sort-of-string-not-array'
 'properties': ...}
```

awm33:
Auditing some existing tap schemas, and testing with a subset of that would give us a better idea of the breadth of cases the target will encounter.

AlexanderMann:
PR for `defaults` in JSON Schema: https://github.com/datamill-co/target-postgres/pull/31

AlexanderMann:
Empty strings are not possible to be replicated, and for columns which are non null strings, they cause the replication to break.

According to the [PSQL docs](https://www.postgresql.org/docs/9.0/sql-copy.html), we have a couple options:
- detect empty string, throw an intelligible error for users
- detect a control string from user input stream (even let them configure it) and use `NULL 'some-string'` in our copy command
  - if detected control string, throw intelligible error
- `FORCE_QUOTE` will force us to quote everything which is not to be interpreted as a null value. This gets around this issue _I think_ in the most straightforward/elegant manner

AlexanderMann:
Okay, so I've got something working with `FORCE_NOT_NULL`. _Unfortunately_, what it ends up doing is making our data error even more specific:

**On columns which _are_ nullable, where you are trying to persist the empty string, `null` will be persisted instead...**

Given the above limitations, I think it might be more reasonable to detect the control `NULL`, and then Error on that. Allow users to configure the control string, and move on. It makes for much better default behaviour.

awm33:
@AlexanderMann I think we can:

1) Set the NULL identifier in copy to "NULL" and detect nulls on the incoming data. WARN on NULL in the incoming stream (often "NULL" is meant to be NULL, so we probably don't want to error)
2) If we want a robust solution. Generate a random 128 bit string, and use that as the NULL identifier in the COPY command.

If 2 seems like it will take a lot more time than 1, we can just go with 1.

awm33:
[...]

PR #17, opened 2018-11-06
Feature/json schema ref support

AlexanderMann:
# Motivation
This tackles a couple things:
- https://github.com/datamill-co/target-postgres/issues/7
- https://github.com/datamill-co/target-postgres/issues/13

Additionally, this does a lite bit of refactoring, moving pure functions over to the `json_schema` file structure.

## Suggested Musical Pairing
https://soundcloud.com/sylvanesso/radio

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,17 +39,22 @@ It ignores "STATE" type Singer messages.
 - Requires a [JSON Schema](https://json-schema.org/) for every stream.
 - Only string, string with date-time format, integer, number, boolean, object, and array types with or without null are supported. Arrays can have any of the other types listed, including objects as types within items. 
     - Example of JSON Schema types that work
-        - ['number']
-        - ['string']
-        - ['string', 'null']
+        - `['number']`
+        - `['string']`
+        - `['string', 'null']`
     - Exmaple of JSON Schema types that **DO NOT** work
-        - ['string', 'integer']
-        - ['integer', 'number']
-        - ['any']
-        - ['null']
+        - `['string', 'integer']`
+        - `['integer', 'number']`
+        - `['any']`
+        - `['null']`
 - Types cannot change
-- JSON Schema combinations such as "anyOf" and "allOf" are not supported.
-- JSON Schema $ref is not supported.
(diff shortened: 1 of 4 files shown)


target_postgres/json_schema.py
```diff
@@ -1,12 +1,186 @@
...
+
+def get_ref(schema, ref):
+    """
+    Given a JSON Schema dict, and a valid ref (`$ref`), get the JSON Schema from within schema
+    :param schema: dict, JSON Schema
+    :param ref: string
+    :return: dict, JSON Schema
+    """
+
+    return _get_ref(schema,
```
awm33: Should `id` based refs be supported? https://json-schema.org/understanding-json-schema/structuring.html#using-id-with-ref

README.md
```diff
@@ -39,33 +39,58 @@ It ignores "STATE" type Singer messages.
...
 - Types cannot change
-- JSON Schema combinations such as "anyOf" and "allOf" are not supported.
-- JSON Schema $ref is not supported.
+- JSON Schema combinations such as `anyOf` and `allOf` are not supported.
+- JSON Schema $ref is partially supported:
+  - ***NOTE:*** The following limitations are known to **NOT** fail gracefully
+  - Presently you cannot have any circular or recursive `$ref`s
+  - `$ref`s must be present within the schema:
+    - URI's do not work
+    - if the `$ref` is broken, the behaviour is considered unexpected
```
awm33: If we add the test case and an exception message, I think we can remove this line.

AlexanderMann:
Looking more into `$id`, I'd vote doing that in a totally separate pr. Reasons being:
- `$id` _also_ defines relative/absolute paths for `$ref`s, so supporting them would mean needing to implement a _much_ smarter path traversal alg
  - this is to say that _even_ if we supported `$id` for `$ref` to point to, we'd prolly need to still limit `$ref`'s formats so that we don't have to deal with more complicated path traversal etc.
- `$id` gets into the whole URI debacle as well

Given the above, I think it's reasonable to put `$id` into the known limitations list for the time being.

PR #22, opened 2018-11-06
Add Singer data collection and sponsorship info

awm33:


awm33:
Trying to get this through quickly just to submit the target to Singer.

Issue #20, opened 2018-11-06
Validate incoming JSON Schema


awm33:
Validate incoming JSON Schema using a JSON Schema python library. Log a warning or throw an exception if the JSON Schema is not valid.

PR #23, opened 2018-11-08
Feature/validate json schema

AlexanderMann:
# Motivation
https://github.com/datamill-co/target-postgres/issues/20

The user experience for a broken JSON Schema is fairly poor. This PR seeks to improve the overall experience for the end user by detecting broken/invalid schemas and aborting immediately instead of trying to swallow exceptions etc.

## Suggested Musical Pairing
https://soundcloud.com/talkingheads/stay-up-late-remastered?in=talkingheads/sets/once-in-a-lifetime-the-talking

### DIFF
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -137,6 +137,17 @@ def simplify(schema):
     return _helper_simplify(schema, schema)
 
 
+def validation_errors(schema):
+    """
+    Given a dict, returns any known JSON Schema validation errors. If there are none,
+    implies that the dict is a valid JSON Schema.
+    :param schema: dict
+    :return: [String, ...]
+    """
+
+    return []
+
+
 def from_sql(sql_type, nullable):
     _format = None
     if sql_type == 'timestamp with time zone':
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -175,6 +175,9 @@ def validation_errors(schema):
 
     errors = []
 
+    if not isinstance(schema, dict):
+        errors.append('Parameter `schema` is not a dict, instead found: {}'.format(type(schema)))
+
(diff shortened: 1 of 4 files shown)


target_postgres/json_schema.py
```diff
@@ -1,5 +1,14 @@
 import re
 
+from jsonschema import Draft4Validator
+from jsonschema.exceptions import SchemaError
+
+
+class Error(Exception):
```
awm33: Can we call this something like `JSONSchemaException`? `Error` is very generic.

target_postgres/json_schema.py
```diff
@@ -99,7 +108,7 @@ def _helper_simplify(root_schema, child_schema):
         try:
             return _helper_simplify(root_schema, get_ref(root_schema, child_schema['$ref']))
         except RecursionError:
-            raise Exception('Target `$ref` "{}" is recursive'.format(get_ref(root_schema, child_schema['$ref'])))
+            raise Error('Target `$ref` "{}" is recursive'.format(get_ref(root_schema, child_schema['$ref'])))
```
awm33: Given the name of the repo/project, the word "Target" here may be confusing. Maybe '`$ref` path "{}" is recursive'?

target_postgres/json_schema.py
AlexanderMann: I'm up for that. I tend to prefer not double namespacing vars (ie, the namespace is _already_ `json_schema` so to get to this you already have to do `json_schema.Error`).

This will be a common pattern in the onslaught of prs I have up, I'll tidy this up.

awm33: @AlexanderMann It's just more common in python to have a distinct name, it also makes the stack trace clearer

Issue #21, opened 2018-11-06
Optionally validate data using JSON Schema


awm33:
As records stream in, validate them against their JSON Schema using a python JSON Schema library.

This validation should be optional and turned on/off via CLI flag or config JSON.

We may want to optionally allow for a certain number of records that would error and reject, similar to Redshift loading. ex allow for 5 erroneous documents.

PR #24, opened 2018-11-12
Feature/validate streamed data against known json schema

AlexanderMann:
# Motivation
https://github.com/datamill-co/target-postgres/issues/21

## Notes
Attempting to _persist_ a record even though validation failed is not something I worked on in this pr. I initially attempted to provide this level of configuration, however, this introduces a number of...fun...intricacies.

For the time being, records which _do not_ meet validation standards are placed into (effectively) a dead letter queue, whose size is configurable.

An open question is on whether we should be flushing those records in the same manner as the other buffers (as they too can fill up and cause problems for memory etc.). I'm awaiting input from a reviewer as to what to do in this vein.

## Suggested Musical Pairing
https://soundcloud.com/talkingheads/take-me-to-the-river-live?in=talkingheads/sets/stop-making-sense-2

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,16 @@ Run `target-postgres` against a [Singer](https://singer.io) stream.
 
 It ignores "STATE" type Singer messages.
 
+### Configuration
+#### Invalid Records
+- `invalid_records_detect`: Optional `Boolean`
+  - Defaults to `true`
+  - Include `false` in your config to disable `target-postgres` from crashing on invalid records
+- `invalid_records_threshold`: Optional `Integer`
+  - Defaults to `0`
+  - Include a positive value `n` in your config to allow for `target-postgres` to encounter at most `n` invalid records
+    per stream before giving up.
+
 ## Known Limitations
 - Requires a [JSON Schema](https://json-schema.org/) for every stream.
 - Only string, string with date-time format, integer, number, boolean, object, and array types with or without null are supported. Arrays can have any of the other types listed, including objects as types within items. 
--- a/README.md
+++ b/README.md
@@ -14,12 +14,11 @@ A [Singer](https://singer.io/) postgres target, for use with Singer streams gene
 
 ## Usage
 
-Create a config file with postgres connection information and target postgres schema.
+Create a [JSON config file](#configjson) similar to the following:
 
 ```json
 {
   "postgres_host": "locahost",
-  "postgres_port": 5432,
   "postgres_database": "my_analytics",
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -35,6 +35,16 @@ Run `target-postgres` against a [Singer](https://singer.io) stream.
 
 It ignores "STATE" type Singer messages.
 
+### Configuration
```
awm33: Can we either remove this header and bump "Invalid records" up a level, or create a true configuration section with a table of all configuration file key/value options?

target_postgres/__init__.py
```diff
@@ -30,7 +30,15 @@ def flush_streams(streams, target, force=False):
         if force or stream_buffer.buffer_full:
             flush_stream(target, stream_buffer)
 
-def line_handler(streams, target, max_batch_rows, max_batch_size, line):
+def report_invalid_records(streams):
+    for stream_buffer in streams.values():
+        if stream_buffer.peek_invalid_records():
+            LOGGER.error("Invalid records detected for stream {}: {}".format(
```
awm33: Since this will run every time, we may want to warn if less than invalid_records_threshold and error if at or greater than invalid_records_threshold

README.md
AlexanderMann: Sounds good. I'd personally lean towards the table. Should we use JSON Schema syntax for the types etc.?

README.md
```diff
@@ -14,26 +14,24 @@ A [Singer](https://singer.io/) postgres target, for use with Singer streams gene
...
-  "postgres_password": "1234",
-  "postgres_schema": "mytapname"
-}
-```
-
-Run `target-postgres` against a [Singer](https://singer.io) stream.
-
+Create a JSON config file with the following parameters, and then run `target-postgres`
+against a [Singer](https://singer.io) stream:
+
```
awm33: Can we keep the example JSON? We can keep it simple with only the most common keys (host, database, username, password). We can also note it's an example.

README.md
```diff
@@ -14,26 +14,24 @@ A [Singer](https://singer.io/) postgres target, for use with Singer streams gene
...
+                                                                  
 ```sh
-	tap-something | target-postgres --config config.json
+  tap-something | target-postgres --config config.json
 ```
-
-It ignores "STATE" type Singer messages.
+NOTE: It ignores "STATE" type Singer messages.postgres connection information and target postgres schema.
+
+| Field | Type | Default | Details |
```
awm33: Love this!

Issue #18, opened 2018-11-06
Column name normalization and error handling


awm33:
Currently, target-postgres blindly uses the JSON Schema property name as the column name https://github.com/datamill-co/target-postgres/blob/master/target_postgres/postgres.py#L597

However, postgres (and redshift) have limitations on what characters can be used in column names, and the length of a column name.

The target should attempt to normalize and shorten a column, and detect conflicts with other columns after doing so, throwing an exception if a conflict us detected.

From https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
```
SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. The SQL standard will not define a key word that contains digits or starts or ends with an underscore, so identifiers of this form are safe against possible conflict with future extensions of the standard.

The system uses no more than NAMEDATALEN-1 bytes of an identifier; longer names can be written in commands, but they will be truncated. By default, NAMEDATALEN is 64 so the maximum identifier length is 63 bytes. If this limit is problematic, it can be raised by changing the NAMEDATALEN constant in src/include/pg_config_manual.h.
```

AlexanderMann:
@awm33 to review how you're thinking this would proceed:
1. given a JSON Schema, translate it to SQL
2. For each column name, normalize*
3. Detect when normalized names will collide and provide details to user by referencing both the original JSON Schema values/paths, and the resulting SQL names
4. Upon pulling the SQL Schema from the target DB, alert user when a _normalized_ name will collide with what is present in the DB
  - this one is also a concern for me as we _cannot_ guarantee that the column which is present in the DB is the same as the one we have now normalized
  - to get around the above, I could see us:
    - providing some configuration options
    - if the target colunm's _type_ is the same as the one we have selected for the JSON Schema, we could potentially warn, and move on

*normalize:
- Truncate name to target DB's standards (detect either `NAMEDATALEN-1` or just hardcode for the time being)
- Shorten: ***??? (this is the part I could actually use some help with as far as what you're thinking)***

-------

If my above understanding is correct about everything I think I can easily get started on the duplicate collision in JSON Schema work. I also think that the refactoring necessary for this will result in a lot of benefits for:
[...]

Issue #4, opened 2018-10-29
Decision - When / how to split off Redshift


awm33:
When is the best time to split off the codebase into:
- datamill-co/singer-sql-target-tools (better names welcome)
- datamill-co/target-redshift

A lot of the issues / features to be worked on overlap, but may require branching logic based on posgres vs redshift.

Is it best to do this after more work is done on postgres?

How best to do this?

AlexanderMann:
When we do eventually do this, I'd recommend step 1 for a _pseudo_ poor implementation to be simply adding support for PSQL 8.0.2 into the postgres branch, then splitting: https://docs.aws.amazon.com/redshift/latest/dg/c_redshift-and-postgres-sql.html

Obviously things like copying via s3 etc. would be...dramatically better...but I think that's for later.

Side note:
- is s3 _really_ just a different tap?

awm33:
I think the when would ideally be in the next 1.5/2 weeks.

The how is roughly:
- Create a `target-sql-tools` repo
- Create a `target-redshift` repo
- Create a shared target class like `TargetSQL` (see redshift branch)
- Refactor / cleanup the code as it's moved to `TargetSQL`. Testing the databases differences as we go. (If we eventually target MySQL or Snowflake, we may need to break the shared functions down further, get separate generated SQL.)
- Reduce the `target-postgres` repo to postgres specific code, referencing the shared `target-sql-tools` library

Some of the above was started in the redshift branch with a rough working version.

awm33:
@AlexanderMann ^

AlexanderMann:
The timeframe of 2 weeks sounds fine to me. I think in order to get there/get ready for there we should potentially do something like the following:

- publish `target-postgres` on `PyPi`
  - [CD for this might be potentially quite useful](https://github.com/datamill-co/target-postgres/issues/6)
- create a `target-sql` _module_ inside of the `target-postgres` repo. Target this as another `module` in `setup.py`
- pull the `target-redshift` into the main `target-postgres` branch as another `module`, similarly target `target-sql`
  - Get testing and CI setup for this
- At the 2 week mark assess if further splitting things into distinct repos (most likely yes)

awm33:
@AlexanderMann 

`PyPi` is not necessary for it to be included by Singer.io and many taps/targets are not published there. I'd like to get it there, but it doesn't seem like the first step or even a hard requirement to splitting the repos.

[...]

Issue #3, opened 2018-10-29
Setup continuous integration


awm33:
Setup continuous integration using CircleCI, TravisCI, or another cloud hosted CI platform.

AlexanderMann:
https://github.com/datamill-co/target-postgres/pull/8

PR #1, opened 2018-10-16
Readme and target schema config

kevinsanz93:
I updated the readme to account for the config values. Also added a config.get for the target catalog

PR #28, opened 2018-11-13
Feature/canonicalize column names

AlexanderMann:
# Motivation
https://github.com/datamill-co/target-postgres/issues/18

This is the first pr in a series. Proper handling of PSQL names involves non-latin characters, unicode, and general madness. As such, I've tried to split this work up into the following passes:

- detect invalid names (ASCII only)
- normalize names (leverage PSQL comments to store information about schema etc.)
- quotes support (?)
- unicode support (?)

## Suggested Musical Pairing
https://soundcloud.com/starfucker_usa/golden-light-1

### DIFF
--- a/target_postgres/postgres_schema.py
+++ b/target_postgres/postgres_schema.py
@@ -1,4 +1,15 @@
+from copy import deepcopy
 import re
+
+from target_postgres.singer_stream import (
+    SINGER_RECEIVED_AT,
+    SINGER_BATCHED_AT,
+    SINGER_SEQUENCE,
+    SINGER_TABLE_VERSION,
+    SINGER_PK,
+    SINGER_SOURCE_PK_PREFIX,
+    SINGER_LEVEL
+)
 
 """
 NAMEDATALEN _defaults_ to 64 in PostgreSQL. The maxmimum length for an identifier is
@@ -43,3 +54,37 @@ def canonicalize_column_name(column_name):
 
     lowered_name = column_name.lower()
     return lowered_name[0] + re.sub(r'[^\w\d_$]', '_', lowered_name[1:])
+
+
+def add_singer_columns(schema, key_properties):
+    ret_schema = deepcopy(schema)
+    properties = ret_schema['properties']
+
+    if SINGER_RECEIVED_AT not in properties:
+        properties[SINGER_RECEIVED_AT] = {
+            'type': ['null', 'string'],
+            'format': 'date-time'
+        }
+
+    if SINGER_SEQUENCE not in properties:
+        properties[SINGER_SEQUENCE] = {
+            'type': ['null', 'integer']
+        }
+
+    if SINGER_TABLE_VERSION not in properties:
+        properties[SINGER_TABLE_VERSION] = {
+            'type': ['null', 'integer']
+        }
+
+    if SINGER_BATCHED_AT not in properties:
+        properties[SINGER_BATCHED_AT] = {
+            'type': ['null', 'string'],
+            'format': 'date-time'
+        }
+
+    if len(key_properties) == 0:
+        properties[SINGER_PK] = {
+            'type': ['string']
+        }
+
+    return ret_schema
--- a/target_postgres/postgres_schema.py
+++ b/target_postgres/postgres_schema.py
@@ -29,12 +29,12 @@ def canonicalize_column_name(column_name):
 
     if not re.match(r'^[a-zA-Z_]', column_name):
         raise SchemaError(
-            'Field {} cannot be canonicalized. Must start with an letter, or underscore'.format(
+            'Field "{}" cannot be canonicalized. Must start with an letter, or underscore'.format(
                 column_name))
 
     if len(column_name) > NAMEDATALEN:
         raise SchemaError(
-            'Field {} cannot be canonicalized. Length {} must be less than or equal to {}'.format(
+            'Field "{}" cannot be canonicalized. Length {} must be less than or equal to {}'.format(
                 column_name,
                 len(column_name),
                 NAMEDATALEN))
--- a/target_postgres/postgres_schema.py
+++ b/target_postgres/postgres_schema.py
@@ -1,0 +1,45 @@
+import re
+
+"""
+NAMEDATALEN _defaults_ to 64 in PostgreSQL. The maxmimum length for an identifier is
+  NAMEDATALEN - 1.
+
+  TODO: Figure out way to `SELECT` value from commands
+"""
+NAMEDATALEN = 63
+
+
+class SchemaError(Exception):
+    """
+    Raise this when there is an error with regards to Postgres Schemas
+    """
+
+
+def canonicalize_column_name(column_name):
+    """
+    Given a column name, canonicalize name such that it can be
+    used as a Postgres column identifier.
+
+    For details or a full list of restrictions, see: https://www.postgresql.org/docs/9.4/sql-syntax-lexical.html
+
+    TODO: allow for non latin characters
+    :param column_name: String
+    :return: String
+    """
+
+    if not re.match(r'^[a-zA-Z_]', column_name):
+        raise SchemaError(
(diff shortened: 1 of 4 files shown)


AlexanderMann:
@awm33 before I went ahead and did it...I figured I'd get your input. I'm thinking that all of the denesting of schema stuffs could be moved out either to `json_schema` or to `postgres_schema`. It'd greatly clean up the main `postgres` file.

target_postgres/postgres_schema.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+NAMEDATALEN _defaults_ to 64 in PostgreSQL. The maxmimum length for an identifier is
+  NAMEDATALEN - 1.
+
+  TODO: Figure out way to `SELECT` value from commands
+"""
+NAMEDATALEN = 63
+
+
+class SchemaError(Exception):
```
awm33: Maybe `PostgresSchemaError` would be more clear in a stack trace?

target_postgres/postgres_schema.py
```diff
@@ -0,0 +1,90 @@
...
+                len(column_name),
+                NAMEDATALEN))
+
+    ## Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($)
+
+    lowered_name = column_name.lower()
+    return lowered_name[0] + re.sub(r'[^\w\d_$]', '_', lowered_name[1:])
+
+
+def add_singer_columns(schema, key_properties):
```
awm33: This function doesn't have any postgres specific code.

awm33 (changes requested):

awm33:
@AlexanderMann I'm not sure if it's the right direction. I'm not a huge OO person, but creating a SQL target OO interface seems like the right direction and this PR is moving away from that. Clojure doesn't really have OO, but you could think of this as developing a protocol for SQL-like targets.

AlexanderMann:
Makes sense. For what it's worth, I have been wanting to head in that
direction for a while, but have also wanted to not slow down velocity on
the task at hand for pure refactors etc.

I'm happy to do the legwork in this pr or close this and do that first.
Thoughts?

bender-peliqan (commented):
## Code Review by Bender (automated)

Adds DELETERECORD support to target-postgres: `target_tools._line_handler` routes the new message type to `StreamTracker.handle_delete_record_message`, which buffers deletes per stream and flushes them **after** the record batch (so upsert+delete of the same PK in one run correctly ends up deleted). The actual DELETE is built as `DELETE FROM schema.table WHERE (pk_cols) IN ((%s,..),(%s,..),...)`, batched at `DELETE_BATCH_SIZE=1000`, with source-cased PK keys canonicalized to physical column names, deterministic skips (missing table / missing PK column) that rollback + return 0, and a `write_batch`-style transient-connection retry loop that raises on real failures. Overall: clean design, good separation, thoughtful test coverage.

### Looks good (verified)
- SQL is fully parameterised (`sql.Identifier`, `sql.Placeholder`); no interpolation of user data.
- Ordering guarantee (records flushed before deletes) is enforced in both `flush_stream` and `flush_streams`, and asserted in `test_records_flushed_before_deletes`.
- Watermark bookkeeping for deletes mirrors record semantics (`stream_add_watermarks`/`stream_flush_watermarks`), so STATE won't emit past an un-flushed delete.
- Canonicalisation of PK keys (`ID` -> `id`, `Account Id` -> `account_id`) closes a real silent no-op path — covered by `test_delete_key_is_canonicalized_to_physical_column` and `test_delete_key_with_invalid_chars_canonicalized`.
- Deterministic skips (missing table, missing PK column) rollback and return 0 rather than raising — correct best-effort behavior; both covered.
- Transient-connection retry loop raises on exhaustion instead of returning 0 (which would otherwise advance the delete bookmark past unlanded deletes) — asserted in `test_connection_loss_retries_then_raises`.
- `_flush_deletes` uses `hasattr(self.target, 'delete_records')` fallback — safe for any non-Postgres target reusing StreamTracker.
- Tests use the real `assert` keyword throughout, and each test has a `# WHY:` comment stating the invariant it protects.

### Concerns / questions

1. **PK-key uniformity assumption.** In `delete_records`:
   ```python
   raw_pk_columns = list(records[0].keys())
   ...
   params = [record[c] for record in batch for c in raw_pk_columns]
   ```
   The PK column set (and iteration order) is taken from `records[0]` and applied to every record in the batch. If a later record has a different key set or ordering (e.g. tap emits `{id, _sdc_division}` for one row and `{_sdc_division, id}` for the next), the second row either KeyErrors — raised as `PostgresError`, aborting the whole batch — or, worse, if the key set matches but the order is different, binds values to the wrong columns and silently deletes the wrong rows. In practice a well-behaved tap emits a stable PK shape per stream, but a defensive assert (`all(set(r.keys()) == set(raw_pk_columns) for r in batch)`) would turn a subtle data-integrity failure into a loud one. Worth adding, or at least a comment stating the invariant the caller must honour.

2. **Partial-PK-missing skips the entire batch.** In the missing-column branch, if the delete records carry keys `[id, division]` and only `id` exists on the table, the whole batch is silently skipped (log warning + return 0). That's the safe choice from a data-loss standpoint, but it means a mild schema drift on a *non-PK* helper column drops deletes on the floor. The log line is clear about *which* column is missing — good — but consider whether this should be an error rather than a skip when the missing column is not itself a PK. Judgement call; documenting the intended semantic in the docstring would be enough.

3. **DELETERECORD for an un-synced stream via `add_table_mapping_helper`.** `test_delete_for_unsynced_stream_is_routed` verifies the routing but stubs the DB entirely. `setup_table_mapping_cache(cur)` presumably repopulates the mapping from the live DB, so a table that exists but was never registered in this process should resolve — but I'd like to see this exercised in the live `test_postgres.py` suite (a delete-only sync of a pre-existing table) before rolling into a dedicated delete-sync run.

4. **`_flush_deletes` bumps the watermark even when the target doesn't implement `delete_records`.**
   ```python
   if hasattr(self.target, 'delete_records'):
       self.target.delete_records(stream, records)
   else:
       LOGGER.warning("Target does not implement delete_records; ...")
   self.stream_flush_watermarks[stream] = self.stream_add_watermarks.get(...)
   ```
   The watermark advance is unconditional. For target-postgres this branch is dead, but any other target reusing `StreamTracker` (target-bigquery, per the companion PR) will silently drop deletes AND emit STATE past them. Consider guarding the watermark bump on the `if` branch, or moving it inside — otherwise the "safe fallback" quietly corrupts the delete bookmark.

### Test coverage

Coverage is thorough at the unit level — routing, ordering, batching, canonicalisation, composite PK, empty input, missing table, missing PK column, execute failure (raises), transient-connection retry (exhaustion + recovery). Two follow-ups worth considering:

- A test with a batch whose second record has a different key set (asserts either a clean raise or the invariant — see concern #1).
- Live-DB coverage in `test_postgres.py` for the un-synced-stream path (concern #3).

### Minor
- `postgres.py` uses `##` (double comment marker) for the `DELETE_BATCH_SIZE` class-level comment — the rest of the file uses `#`. Cosmetic.
- `deleted_total += cur.rowcount if cur.rowcount and cur.rowcount > 0 else 0` — the `cur.rowcount and cur.rowcount > 0` is redundant; `max(cur.rowcount or 0, 0)` reads cleaner. Nit.
- Docstring on `delete_records` is unusually good (retry semantics, skip semantics, canonicalisation rationale all called out). Keep this bar.

### Recommendation
Comment / approve-with-suggestions. The core delete pipeline is well-designed, the SQL layer is parameterised, and the tests genuinely protect the invariants that matter (ordering, canonicalisation, no-silent-swallowing-of-real-failures). Nothing here is blocking, but I'd address concern #4 (unconditional watermark advance in the no-delete-support branch) before this ships alongside the target-bigquery companion PR — that's the branch most likely to bite in practice.


---
_Warning: no reviewers are currently assigned to this PR._

@AlexanderMann FYI — automated review posted.

PR #31, opened 2018-11-14
Fix/null default

AlexanderMann:
# Motivation
JSONSchema defines a `default` field to hint to validators, etc., that the given property does not need to meet it's `not null` constraint, and that if this failure case is reached, to simply use the default instead.

Presently, we do not handle a couple things:
- create a table with defaults
- our persist method doesn't play nice with default values in general
  - even if the column has a default set in SQL, it will not utilize that when the copy places `null`
  - this becomes especially tricky when you have mixed null/not-null values in a csv column...

## Notes
If there is a simple solution to get the CSV copy to play nicely, then I think we should use that.

This brute force approach works though.

## Suggested Musical Pairing
https://soundcloud.com/ugly-casanova/barnacles

### DIFF
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -101,12 +101,14 @@ def is_nullable(schema):
 
 
 def _helper_simplify(root_schema, child_schema):
+    ret_schema = {}
+
     if not child_schema:
-        return child_schema
+        pass
 
     elif is_ref(child_schema):
         try:
-            return _helper_simplify(root_schema, get_ref(root_schema, child_schema['$ref']))
+            ret_schema = _helper_simplify(root_schema, get_ref(root_schema, child_schema['$ref']))
         except RecursionError:
             raise JSONSchemaError('`$ref` path "{}" is recursive'.format(get_ref(root_schema, child_schema['$ref'])))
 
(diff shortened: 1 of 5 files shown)


target_postgres/json_schema.py
```diff
@@ -101,12 +101,14 @@ def is_nullable(schema):
 
 
 def _helper_simplify(root_schema, child_schema):
+    ret_schema = {}
+
     if not child_schema:
-        return child_schema
+        pass
```
awm33: Could the if just be removed here then? Does it really not affect thel ogic?

tests/fixtures.py
```diff
@@ -36,6 +36,10 @@
             'name': {
                 'type': ['string']
             },
+            'paw_size': {
+                'type': ['integer'],
+                'default': 314159
```
awm33: That's a big paw

AlexanderMann: These cats have paws for pi and pi alone

target_postgres/json_schema.py
AlexanderMann: Actually, I think that case covered up another bug. Schemas are totally allowed to be completely empty. Those reference objects types. I'll add a better check for that.

tests/fixtures.py
awm33: @AlexanderMann cat pi https://i.pinimg.com/originals/60/46/5d/60465da88007d8e5b28f265c7805abce.jpg

PR #32, opened 2018-11-15
Fix/empty string

AlexanderMann:
# Motivation
https://github.com/datamill-co/target-postgres/issues/7#issuecomment-438861733

## Suggested Musical Pairing
https://soundcloud.com/planned_obsolescence/mixed-bizness

### DIFF
--- a/target_postgres/postgres.py
+++ b/target_postgres/postgres.py
@@ -20,6 +20,9 @@ from target_postgres.singer_stream import (
     SINGER_LEVEL
 )
 
+RESERVED_NULL_DEFAULT = 'NULL'
+
+
 class PostgresError(Exception):
     """
     Raise this when there is an error with regards to Postgres streaming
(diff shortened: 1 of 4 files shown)


target_postgres/postgres.py
```diff
@@ -561,15 +564,30 @@ def transform():
...
-                    for prop in datetime_fields:
-                        if prop in row:
+                        ## Serialize datetime to postgres compatible format
+                        if prop in datetime_fields \
+                                and prop in row:
                             row[prop] = self.get_postgres_datetime(row[prop])
+
+                        ## Serialize NULL default value
+                        if row.get(prop, False) == RESERVED_NULL_DEFAULT:
+                            self.logger.warn('Reserved {} value found at: {}.{}.{}'.format(
```
awm33: I'd add a little more clarification here, that the reserved value will be intepretted as literally NULL.

PR #37, opened 2018-11-17
Refactor/no more sql column defaults

AlexanderMann:
# Motivation

In the wake of working on column type changes, and hence column splitting, I encountered that our handling of sql defaults actually never _quite_ worked as we anticipated.

Tables manually created in PSQL to have various defaults were not correctly deserialized due to the type of the `information_schema.column_default` column to be `character_data`:

```
{'name': 'foo',
 'type': 'TABLE_SCHEMA',
 'schema': {'properties': 
   {'id': {'type': ['integer', 'null']},
    'bar': {'type': ['string'], 'default': "''::text"}}}}
```

After a conversation, @awm33 and I came to the conclusion that "prefilling"/"backfilling" data is actually not really in keeping with the spirit of providing a streaming data tool. The previous data _did not have_ the values we're trying to assign to it, so we're actually breaking history.

## Notes

There was a conversation about making _all columns_ nullable, however I think that's probably overboard for what we want/have to provide at the moment.

I think that if a column _becomes_ nullable, we should probably "expand" the column to handle that, but if your schema has 5 non nullable columns...and never changes, the user experience will be far superior to have the non null constraints in place.

The above is also a smaller changeset. I think if we want to go _further_ down the path of making every column nullable, we should make a pr _just_ for that.

This was my spike pr: https://github.com/datamill-co/target-postgres/compare/master...AlexanderMann:scratchwork/column-type-renames?expand=1

## Suggested Musical Pairing

Wind.


target_postgres/postgres.py
```diff
@@ -622,25 +622,20 @@ def get_postgres_datetime(self, *args):
     def add_column(self, cur, table_schema, table_name, column_name, column_schema):
         data_type = json_schema.to_sql(column_schema)
 
-        default_value = self.get_null_default(column_name, column_schema)
-
-        if default_value is not None:
-            default_value = sql.SQL(' DEFAULT {}').format(sql.Literal(default_value))
-        elif not self.is_table_empty(cur, table_schema, table_name):
-            raise PostgresError('Non-trival default needed on new non-null column `{}.{}.{}`'.format(
+        if not json_schema.is_nullable(column_schema) \
```
awm33: @AlexanderMann Aren't we always forcing columns to be not nullable, regardless if the table is empty?

awm33:
@AlexanderMann Do we want to make the "primary key" columns not nullable? We may also want to not allow for multiple types on PK columns. I think being stricter on those columns is permissible.

AlexanderMann:
@awm33 happy to move forward with every column being nullable. I wrote a blurb in the PR description about why I ended up _not_ doing that.

Really, the decision was twofold:

- smaller changeset for this pr
- better/more expected user experience for schema which _don't_ change
   - to this end, this is totally a thing for the pks

AlexanderMann:
Another thing to note about the PKs, we already do pretty aggressive type/value checking against those here:

- [makes sure that the `key_properties` _themselves_ are the same](https://github.com/datamill-co/target-postgres/blob/master/target_postgres/postgres.py#L76-L82)
- [makes sure that the types of each individual key prop have not changed](https://github.com/datamill-co/target-postgres/blob/master/target_postgres/postgres.py#L743-L744)

ie, the above will fail if the pks change OR their types change.

Issue #12, opened 2018-11-01
Types cannot change / mixed types


awm33:
If a column type changes, the target should be able to handle that. There is no Singer standard for handling this, so we could look to Stitches standard. Though I think we can use JSON Schema to our advantage, since at least for now, we are assuming it is always available, and not handling any given JSON.

I think this comes down to:
- Identify a set of rules for handling data type changes. Try to maximize user experience, and minimize negative consequences of changing the schema.
- Implement those rules and create tests around each combination of type change cases.

AlexanderMann:
Following the Stitch guidelines would be involved, but probably fairly rewarding. The way they handle rejected records is pretty good: https://www.stitchdata.com/docs/destinations/postgresql/#rejected-records-log

Presently we don't do type suffixes on column names (...right?). Looks like Stitch only does that on the [first detected difference](https://www.stitchdata.com/docs/data-structure/postgresql-data-loading-behavior#stitch-detects-multiple-data-types-for-a-single-colum). To that end, that'll effect the [conversation here](https://github.com/datamill-co/target-postgres/issues/18) somewhat.

I think following Stitch's example for how to handle this makes the most sense.

---------

In general I think it might make a lot of sense to start using Stitch's documentation as the specification for features added to both PostgreSQL and Redshift: https://www.stitchdata.com/docs/data-structure/postgresql-data-loading-behavior

awm33:
@AlexanderMann The biggest reason Stitch added the rejected records was because they had no way to expose loading logs to customers at the time. We may want to consider adding a rejected logs table (I actually think it was inspired by Redshift's built-in table that is similar for COPY errors), but we could start with logging.

> Presently we don't do type suffixes on column names
No, but we could. Yes, it will effecting column naming.

awm33:
@AlexanderMann That all sounds right, except for:
- With postgres we can just keep `string` mapped to `text` types.
- Redshift will need varchar expansion.
- For integers, do we just want to make all integers 64bit? There are table size implications there, but for practical purposes, negligible.

AlexanderMann:
@awm33 sounds like widening is a feature to me! No sense in trying to do that yet if there's no pressing demand for it.

Once we get to Redshift, there'll be an obvious performance pr...but that's a battle for future-us.

awm33:
@AlexanderMann Agreed. Redshift is the one with the real need for widening.

AlexanderMann:
(sorry, accidentally pressed close!)

PR #38, opened 2018-11-19
Feature/column type splitting

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/12

Presently if a column comes to us with a type change, we are unable to handle the type change. The work done in this pr:

- detects the type change
- creates two new columns:
  - `<original-column-name>__<existing-type:sql-type-shortand>`
  - `<original-column-name>__<new-type:sql-type-shortand>`
- migrates the data from `<original-column-name>` to `<original-column-name>__<existing-type:sql-type-shortand>`
- drops `<original-column-name>`

## Notes

- This lays some groundwork for future `$any` support
- This lays some groundwork for future `regular-type -> null-type` support but does not implement the work to do this

## Suggested Musical Pairing

Dog sleeping.

### DIFF
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -279,3 +279,32 @@ def to_sql(schema):
         sql_type += ' NOT NULL'
 
     return sql_type
+
+
+_shorthand_mapping = {
+    'null': '',
+    'string': 's',
+    'number': 'f',
+    'integer': 'i',
+    'boolean': 'b',
+}
+
+
+def _type_shorthand(type_s):
+    if isinstance(type_s, list):
+        shorthand = ''
+        for type in sorted(type_s):
+            shorthand += _type_shorthand(type)
+        return shorthand
+
+    if not type_s in _shorthand_mapping:
+        raise JSONSchemaError('Shorthand not available for type {}. Expected one of {}'.format(
+            type_s,
+            list(_shorthand_mapping.keys())
+        ))
+
+    return _shorthand_mapping[type_s]
+
+
+def sql_shorthand(schema):
+    return _type_shorthand(get_type(schema))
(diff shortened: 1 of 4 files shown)


target_postgres/json_schema.py
```diff
@@ -279,3 +279,32 @@ def to_sql(schema):
         sql_type += ' NOT NULL'
 
     return sql_type
+
+
+_shorthand_mapping = {
```
awm33: @AlexanderMann Stitch uses full type names https://www.stitchdata.com/docs/data-structure/table-structural-changes#columns-mixed-data-types

target_postgres/postgres.py
```diff
@@ -543,20 +542,48 @@ def get_update_sql(self, target_table_name, temp_table_name, key_properties, sub
                         insert_distinct_on=insert_distinct_on,
                         insert_distinct_order_by=insert_distinct_order_by)
 
+    def serialized_value(self, field, value, default_fields, datetime_fields):
```
awm33: @AlexanderMann I don't think we want to combine defaulting and serialization (eg formatting dates). Those seem like distinct things to me. I think creating true functionality around serialization, which in this case is a CSV, may make sense. As in, call a serialize function per row that returns a CSV row, and serialize each field as needed.

target_postgres/postgres.py
```diff
@@ -582,36 +559,51 @@ def persist_rows(self,
...
-
-                            row[mapped_field] = self.serialized_value(prop,
-                                                                      row.get(prop, None),
-                                                                      default_fields,
-                                                                      datetime_fields)
-                            row.pop(prop, None)
+                record = next(records_iter)
+                row = {}
+
+                for field in set(headers + list(record.keys())):
```
awm33: If we are validating JSON, shouldn't `headers` and `record.keys()` be the same? If there are extra columns in `record.keys()`, wouldn't that fail?

AlexanderMann: Column splitting. Headers will contain the mapped fields ie `name__s`, whereas record will contain `name` etc. I can also get this from filtering the mapped field names. Lemme do that instead to make this clearer.

PR #41, opened 2018-11-21
Refactor/rbdms base

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/projects/1#card-14578993
https://github.com/datamill-co/target-postgres/pull/29

In this PR we introduce `RDBMSInterface`. This is aimed at being the base class for generic relational database targets.

The work for building out a full `RDBMSInterface` is loosely scheduled as:
- schema updates/schema processing
- batch denesting/batch writing
- version updates

This PR only aims to start tackling:
- initialization/bones of RDBMSInterface
- schema processing

# Notes

As noted in `rdbms_base`:

```py
# RDBMS Base
## This module is the base implementation for Singer RDBMS target support.
## Expected usage of this module is to create a class representing your given
## RDBMS Target which overrides RDBMSInterface.
#
# Transition
## The given implementation here is in transition as we expand and add various
## targets. As such, there are many private helper functions which are providing
## the real support.
##
## The expectation is that these functions will be added to RDBMSInterface as we
## better understand how to make adding new targets simpler.
#
```

## Suggested Musical Pairing
https://soundcloud.com/santigold/disparate-youth


AlexanderMann:
NOTE: This pr is currently ballooned a little bit as it is branched off of the work done in:
- #39 
- #40 

Once those are merged the overall diff here will be more manageable.

target_postgres/rdbms_base.py
```diff
@@ -0,0 +1,273 @@
...
+        key_prop_schemas[key] = schema['properties'][key]
+    _denest_schema(root_table_name, schema, key_prop_schemas, subtables)
+
+    ret = []
+    for name, schema in subtables.items():
+        ret.append(to_table_schema(name, schema['level'], schema['key_properties'], schema['properties']))
+    return ret
+
+
+class RDBMSNotImplementedError(NotImplementedError):
```
awm33: This seems unnecessary, it's pretty common to just throw straight up NotImplementedError from a base class function.

target_postgres/rdbms_base.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    Raise when a RDBMS method has been left unimplmented which _must_ be overridden
+    in order for the interface to work correctly.
+    """
+
+
+class RDBMSInterface:
```
awm33: I kinda hate RDMS lol. Doesn't roll off the tongue like SQL (sequel).

PR #43, opened 2018-11-26
Refactor/sql persist

AlexanderMann:
# Motivation
https://github.com/datamill-co/target-postgres/projects/1#card-14578993

Follow up to the work done in #41. This PR seeks to start tackling the commonality of `write_batch` for `SQLInterface`.

## Notes
This pr moves the following functionality over to `sql_base`:
- records denesting per table
- records serialization preparation for remote persistence

This pr _removes_ the following functionality from `sql_base`:
- bulk/entire schema updating for remote
  - it leaves in place updating _individual_ tables

## Suggested Musical Pairing
https://soundcloud.com/subpop/rolling-blackouts-julies-place

### DIFF
--- a/target_postgres/postgres.py
+++ b/target_postgres/postgres.py
@@ -115,12 +115,12 @@ class PostgresTarget(SQLInterface):
                                 json_schema.get_type(stream_buffer.schema['properties'][key])
                             ))
 
-                written_batches_details = self.write_table_batches(cur,
-                                                                   root_table_name,
-                                                                   stream_buffer.schema,
-                                                                   stream_buffer.key_properties,
-                                                                   records,
-                                                                   {'version': target_table_version})
+                written_batches_details = self.write_batch_helper(cur,
+                                                                  root_table_name,
+                                                                  stream_buffer.schema,
+                                                                  stream_buffer.key_properties,
+                                                                  records,
+                                                                  {'version': target_table_version})
 
                 cur.execute('COMMIT;')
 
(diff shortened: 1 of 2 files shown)


target_postgres/postgres.py
```diff
@@ -407,93 +306,84 @@ def get_update_sql(self, target_table_name, temp_table_name, key_properties, sub
...
-
-        datetime_fields = [k for k,v in streamed_json_schema['schema']['properties'].items()
-                           if v.get('format') == 'date-time']
-
-        default_fields = {k: v.get('default') for k, v in streamed_json_schema['schema']['properties'].items()
-                          if v.get('default') is not None}
-
-        fields = set(headers +
-                     [v['from'] for k, v in remote_schema.get('mappings', {}).items()])
+    def parse_table_record_serialize_field_name(self, remote_schema, streamed_schema, field, value):
```
awm33: @AlexanderMann You're using the word "parse" a lot. Whenever I see that in code I take it as meaning interpreting / deserializing something from a string. I realize the that's not the only use of the word parse, and context can help, but I think it's losing exact meaning in here.

target_postgres/sql_base.py
```diff
@@ -214,38 +332,212 @@ def get_table_schema(self, connection, name):
...
+        Update the remote table schema based on the merged difference between
+        `remote_table_json_schema` and `table_json_schema`.
         :param connection: remote connection, type left to be determined by implementing class
         :param remote_table_json_schema: get_table_schema
         :param table_json_schema: updates for get_table_schema
-        :param metadata: additional metadata needed to implementing class
+        :param metadata: additional metadata needed by implementing class
         :return: updated_remote_table_json_schema
         """
         raise NotImplementedError('`update_table_schema` not implemented.')
```
awm33: I think it may make sense to start moving these towards the top of the class and eventually make it more obvious that these are the methods to override. Documentation helps too of course

PR #46, opened 2018-11-27
Feature/invalid identifier names

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/18

When we get an identifier for a column/table which is invalid for PostgreSQL we presently don't do anything intelligent and simply try to persist it to the remote target.

This PR aims to do a few things:
- alert users with a fine grained description of the problem by both detecting when there could be a problem and also by detecting when a problem has occurred in the remote and providing details
- map invalid _column_ names to valid names by:
  - lowercasing all names
  - stripping invalid characters and replacing them with `_`
- start building out the usage of `mappings` and exploring commonality of this mechanism
  - my _guess_ is that we will want to make this a universal `SQLInterface` type thing, but for the time being we can leave this alone and keep this a feature specific pr

## Suggested Musical Pairing

https://soundcloud.com/wmi/metronomy-04-love-letters-2?in=metronomy/sets/love-letters-19

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,6 @@ _The above is copied from the [current list of versions](https://www.postgresql.
         - `['integer', 'number']`
         - `['any']`
         - `['null']`
-- Types cannot change
 - JSON Schema combinations such as `anyOf` and `allOf` are not supported.
 - JSON Schema $ref is partially supported:
   - ***NOTE:*** The following limitations are known to **NOT** fail gracefully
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -105,10 +104,18 @@ _The above is copied from the [current list of versions](https://www.postgresql.
     - URI's do not work
     - if the `$ref` is broken, the behaviour is considered unexpected
 - Any values which are the `string` `NULL` will be streamed to PostgreSQL as the literal `null`
+- Table identifiers are restricted to:
```
awm33: Maybe use "Table names ..." here? If you're not used to the nomenclature it may take a second for "identifiers" to click.

README.md
```diff
@@ -105,10 +104,18 @@ _The above is copied from the [current list of versions](https://www.postgresql.
...
+  - 63 characters in length
+  - can only be composed of `_`, lowercase letters, numbers, `$`
+  - cannot start with `$`
+  - ASCII characters
+- Field/Column identifiers are restricted to:
```
awm33: Dido

target_postgres/postgres.py
```diff
@@ -391,24 +416,42 @@ def get_postgres_datetime(self, *args):
...
-            column_name=sql.Identifier(column_name),
+            column_name=sql.Identifier(canonicalized_name),
             data_type=sql.SQL(data_type))
 
-        cur.execute(to_execute)
+        try:
+            cur.execute(to_execute)
+        except Exception as ex:
+            ## Name collision with non mapped, non canonicalized name
+            if column_name != canonicalized_name:
```
awm33: Is there a chance that an exception could be thrown and the statement to be true for reasons other than a name collision? Is there anything in the exception or error message we could inspect?

awm33:
@AlexanderMann Just a couple quick things on the README and a question on error handling

README.md
AlexanderMann: Good call

target_postgres/postgres.py
AlexanderMann: So, in this _very specific_ instance we know:
- an error has been thrown trying to execute a column ***add*** to remote
- the `column_name` we have present in our _streamed_ schema, ***does not*** match the `canonicalized_name` we've generated

In this case, we can safely assume the the likelihood of some collision having happened to be pretty great (ie, I think we can add the context to the end user and it'd be beneficial).

To that point though, I think this can be made _more generic_ and still be beneficial. I'll throw in an exception chain here and modify the message to remove the `if` and we can get the same benefit to the end user.

PR #55, opened 2018-12-04
Feature: init DECISIONS.md

AlexanderMann:
# Motivation

While some of these decisions around more complicated functionality in `target-postgres`/`target-sql` are fresh, write em down.

## Suggested Musical Pairing

https://soundcloud.com/of-montreal/oslo-in-the-summertime

### DIFF
--- a/DECISIONS.md
+++ b/DECISIONS.md
@@ -4,11 +4,13 @@ This document is intended to provide clarity on many of the decisions/rationaliz
 which exist inside of [Datamill's](https://datamill.co/) Target SQL project
 for [Singer](https://singer.io).
 
+## Principles
+
 The guiding principles we try to adhere to herein as far as _how_ to reach a
 conclusion are:
 
 1. When possible, make the resulting data/schema in the remote target consistent, no matter the ordering of potential messages
-  - ie, if our decision would result in a random schema being produced in the remote target for no reasonable benefit, this is in violation
+    - ie, if our decision would result in a random schema being produced in the remote target for no reasonable benefit, this is in violation
 1. Do right by the common _majority_ of users
 1. Make a best effort to prevent a user from having to intervene
 1. Use [Stitch’s offering and documentation](https://www.stitchdata.com/docs) as best practice guidance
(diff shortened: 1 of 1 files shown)


DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+1. Make a best effort to prevent a user from having to intervene
+1. Use [Stitch’s offering and documentation](https://www.stitchdata.com/docs) as best practice guidance
+
+## Schema
+
+### De-nesting
+
+#### What
+
+- [JSON Schema](https://json-schema.org/) allows for complex schemas which have non-literal (ie, compositional) elements
```
awm33: Maybe define `non-literal (ie, compositional) elements` more and give examples? Same may call them basic types, scalar values, etc.

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+- Standard SQL does not support compositional elements, but rather demands something closer to full denormalization
```
awm33: The nature of a relational database is to create a normalized schema of many related tables. It's the best SQL querying experience comes from data that is broken down into related tables. In a way, we are normalizing (even re-normalizing) the structures.

Data warehouses do like denormalized structures while operational databases prefer normalized structures. Here we normalize the structure so the use can choose what to do with the normalized raw data. Also it's easy to access and transform later than JSON blobs.

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+- To overcome this, `target-sql` provides tooling which unpacks json `objects` into their parent record, and json `arrays` as sub tables
```
awm33: Maybe add some small examples of that unpacking below?

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Why
+
+This is the same approach that Stitch Data takes with `array` de-nesting.
```
awm33: The root reason why is end user experience. If we were to place them as JSON/JSONB fields or text blobs, or even through most of them away in the case of redshift. It would be a poor user experience for most users.

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Column Type Mismatch
+
+#### What
+
+- A field has been streamed to the remote target with type `integer`
```
awm33: Love the story approach 👍 . Maybe numbers could be used to make it more obvious it's sequential?

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+- A new field with the _same raw name_ as the remote column has been streamed but has type `BAR`
+- Data of type `BAR` cannot be placed into a column of type `boolean`
+- `target-sql` has tooling which will:
+  - rename the original column to `original_field_name__i`
+  - make the renamed column `nullable`
+  - create a new column of name `original_field_name__b`
+  - stream new data to `original_field_name__b`
+  - (to see a full list of type suffixes, please see: [`json_schema._shorthand_mappings`](https://github.com/datamill-co/target-postgres/blob/master/target_postgres/json_schema.py#L283))
```
awm33: May want to use the most recent commit hash in the URL, otherwise the line number will shift over time.

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+By renaming and migrating the column we:
+
+- make the resulting structure in the database the same no matter whether we upload column `integer` _then_ column `boolean` or vice versa.
+- users learn of dependent views/columns blocking a type change _early_
+
+### Column Name Collision
+
+#### What
+
+- When attempting to `upsert_table`, `SQLInterface` has to handle name
```
awm33: ..collisions.

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+### Objects
+
+#### What
+
+- `Objects` are unpacked into their parent table.
+- The unpacked fields are prefixed with the name of the `field` which originally contained the object.
+
+#### Why
+
+This approach is inspired by what Stitch Data takes with `object` de-nesting.
```
awm33: More of a root why, accessing nested objects as first class fields provides a better user experience.

DECISIONS.md
```diff
@@ -0,0 +1,119 @@
...
+### Arrays
+
+#### What
+
+- `Arrays` are unrolled as individual rows into a child table
+- The table name is constructed as `parent_table__field`
+
+#### Why
+
+This approach is inspired by what Stitch Data takes with `array` de-nesting.
```
awm33: More of a root why, accessing an array/list of nested types as a separate table provides a better user experience.

PR #60, opened 2018-12-07
Feature/table name normalization

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/pull/57

Handles table name normalization:
- invalid names get mapped to valid ones
  - handles truncation
- name collisions are dealt with by:
  - truncating
  - appending an incrementing number till no collision is detected

After a table has been mapped, it continues to persist data to the mapped name.

All data is stored on the `root_table` so that `activate_version` remains simple. This can easily be documented in `RATIONALIZATIONS.md`.

## Notes

DOES NOT handle root table name mapping, but does put in place validation to make sure root table name is able to be persisted.

## Suggested Musical Pairing

https://soundcloud.com/limes-3/saucin-w-h-e-r-b

### DIFF
--- a/target_postgres/postgres.py
+++ b/target_postgres/postgres.py
@@ -72,6 +72,7 @@ class PostgresTarget(SQLInterface):
                     versions.add(record_version)
 
                 current_table_schema = self.get_table_schema(cur,
+                                                             (stream_buffer.stream,),
                                                              stream_buffer.stream)
 
                 current_table_version = None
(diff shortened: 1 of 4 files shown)


target_postgres/postgres.py
```diff
@@ -70,6 +72,7 @@ def write_batch(self, stream_buffer):
                     versions.add(record_version)
 
                 current_table_schema = self.get_table_schema(cur,
+                                                             (stream_buffer.stream,),
```
awm33: @AlexanderMann You can use `[ ]` too for list concatenation, if you want to avoid the dangling `,`

target_postgres/sql_base.py
```diff
@@ -402,6 +394,76 @@ def _canonicalize_identifier(self, name, schema, existing_columns_raw_names, exi
...
+        i = 0
+        ## NAME COLLISION
+        while canonicalized_name in to_from:
+            i += 1
+            suffix = SEPARATOR + str(i)
+            canonicalized_name = raw_canonicalized_name[
+                                 :self.IDENTIFIER_FIELD_LENGTH - len(suffix)] + suffix
+
+            # TODO: logger warn
+            ##raise Exception(
```
awm33: Should this be commented out?

target_postgres/postgres.py
AlexanderMann: @awm33 nope, tuples brah. In Python if you wanna create a tuple you just do `(val,)`.

I'm using them like this here to avoid weirdness as we do things further downstream. ie, if we use tuples for these operations, things _should_ behave nicely and will prevent the implementing class from accidentally modifying something.

target_postgres/sql_base.py
AlexanderMann: Yup. So I think a future refactor (which I have held off on just because it's a #nicetohave) is to move the logger down into `SQLInterface`. Once we do that, we can put back in these logger warns etc.

For this specifically, I think it'd be worthwhile to `warn` the user that a collision has been detected and that we're mucking about with things. The exception that _used_ to be thrown has all of the worthwhile information. It makes more sense in the commit history view of things, but yeah, the TODO is still useful, though I think it'd be more worthwhile as a separate issue etc.

target_postgres/postgres.py
awm33: @AlexanderMann Does this have to do with mutability of a tuple vs list?

AlexanderMann: Yeah, it's that and then the comparison of tuples vs lists is much more straightforward. Either way, we can refactor this in the future if folks find this to be far too confusing etc.

Issue #61, opened 2018-12-07
Stabilize V0 of `target-postgres`


AlexanderMann:
# Motivation

`Target-Postgres` has been changing dramatically for some time. While _many_ of these changes are bug fixes, or straight features, many are modifying the underlying metadata/schema and are not purely "breaking" from a "can I rebase my branch onto this" standpoint.

As such, I'm proposing a new Milestone/Issue to gather up what bugs and housekeeping we want to get in place before going under a semvar versioning guide for ***persisted metadata/schema/data ONLY***. ie, we still will not have to play nice with those who wish to write _code_ on top of our repo, but we _will_ play nice with future versions of our code and we _will_ support the data which is persisted by v0.

## Notes

Please understand that while I am in full support of making life simpler for other devs to work on top of our work, I'm much more concerned about Datamill or someone trying to use this and then getting a tonne of data moved over...which then has to be blown away because we don't know how to migrate etc.

## Outstanding Work

On the top of my mind, the work I think that needs to be done:

- [ ] in flight feature work:
  - [ ] https://github.com/datamill-co/target-postgres/issues/52
- [ ] refactor table mappings and column mappings to use similar/same structure
  - [ ] #44 
- [ ] nested support as noted in:
  - [ ] #42 
  - [ ] #58 
- [ ] Better testing
  - [ ] Full integration test against Github/BigCommerce/whatever in CI (could be low hanging fruit for bugs etc.)

## Suggested Musical Pairing

https://soundcloud.com/hotchip/huarache-lights

AlexanderMann:
@awm33 I believe this to be done!

awm33:
@AlexanderMann Mhmmm

Issue #58, opened 2018-12-06
Denesting arrays assumes items are not literals


awm33:
At some points the code is assuming array items cannot be literals / scalar values. Currently it is failing to denest:

```
{
    ...
    "categories": [23, 22],
    ...
}
```

AlexanderMann:
So this touches on a couple things:

- we need to pick a name for the column under which to put these values in the `root_table_name__categories` table
   - proposal: `root_table_name__categories: __literal`
- nested arrays of arrays are also not handled...which is a whole other bit o' fun. ie:
  - `{... 'categories': [[1, 2], [3, 4], ...] ...}` and more levels of nesting
  - To handle this, I'd propose doing:
    - `root_table_name__categories: __index__0, __index__1 ...` (or something such as)
- Another fun problem: `{type: array,  items: {type: oneOf ...}`
  - in this case, I _think_ we want to _merge_ the resulting tables, not try to postfix them with a type...I think...

AlexanderMann:
https://github.com/datamill-co/target-postgres/pull/63 demonstrates the above.

AlexanderMann:
# Reference
## Nested dictionaries per stitch:

> [object_name]__[property_name]__[sub_propery_name]

## Nested arrays per stitch:

Mentions `_sdc_level_n` column, but does not go in depth about what happens when n

## Literal arrays per stitch:

N/A

# Proposal

## Nested dictionaries

Handle in a similar manner to that provided by Stitch. ie, denest into parent table, make the column name long and ugly.

## Literal Arrays

Separate table. Use similar pks as to how we deal with object arrays currently. Value goes into `__value` column, index goes into `__index` column. (better prefix desired, but don't want to use `sdc` like everything else as this is _not_ a `sdc` thing)

## Nested Arrays

Similar handling to literal arrays. Continue de-nesting values, making long ugly tables with names like:

`root__nested_array_column____value____value____value` etc.

AlexanderMann:
Per Stitch:

> Arrays of literals are treated the same as arrays of objects _except_ the values in the array will be denested into a `value` column` and the index of the value will be placed in the `_sdc_level_<n>_id`

@awm33 I might still propose that we tuck values into somthing like `_sdc_value` etc. (choose your prefix etc.)

AlexanderMann:
@awm33 this pr now has fixes for _some_ of the known bugs: https://github.com/datamill-co/target-postgres/pull/63

[...]

Issue #42, opened 2018-11-26
Nested values throw/are treated differently


AlexanderMann:
# Motivation

When denesting values/records, errors are thrown, and the internal mechanisms for denesting are handling data in inherently different ways.

Presently I'm unsure whether this is _intentional_ or whether this is a 🐛.

#### Nested dictionaries

```py
[{'x0':
    {'x1': {...}}}]
```

Throws invalid number of args exception originating from `denest_subrecord`.

vs

```py
[{'x0':
    {'x1': 123}}]
```

Results in:

```py
{'root_table__x0': [
    {'x1': 123}
  ]}
```

#### Nested nulls

```py
[{'x0':
    {'x1': None}}]
```

Results in:

```py
{'root_table__x0': [
    {'x1': None}
  ]}
```

vs

```py
[{'x0': None}]
```

Results in:

```py
{'root_table': [
    {}
  ]}
```

For this, there appears to be a discrepancy between how `denest_record` and `denest_subrecord` handle `None` values. This is _less_ of a problem and more of a 😕 when looking at the code. Especially because of the comment present only in `denest_record`:

> "## nulls mess up nested objects"

## Suggested Musical Pairing

https://soundcloud.com/tiga/tiga-the-martinez-brothers-aphex-n-girls?in=tiga/sets/tiga-the-martinez-brothers-2

AlexanderMann:
@awm33 yeah, that's the bit. I agree it's a bug, but I'd also like to get a set of failing tests around it etc. in order to resolve it.

ie, this _should be_ very easy to patch, just need to prioritize etc.

AlexanderMann:
https://github.com/datamill-co/target-postgres/pull/63 demonstrates the above.

PR #63, opened 2018-12-08
Fix/nested tables

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/7
https://github.com/datamill-co/target-postgres/issues/42
https://github.com/datamill-co/target-postgres/issues/58
https://github.com/datamill-co/target-postgres/issues/61

Nested table issues. Each known failing problem has been added as a commit.

## Notes

This is pseudo-branched off of https://github.com/datamill-co/target-postgres/pull/60 due to needing more complex `db_cleanup` logic.

As such there _is_ a failing test kicking around in here which is _not_ related to the problems this pr will _ultimately_ be addressing.

## Suggested Musical Pairing

https://soundcloud.com/phoenix/heatwave


target_postgres/json_schema.py
```diff
@@ -4,6 +4,7 @@
 from jsonschema import Draft4Validator
 from jsonschema.exceptions import SchemaError
 
+NULL = 'null'
```
awm33: Should we create one for each JSON Schema type we use?

tests/test_sandbox.py
```diff
@@ -0,0 +1,57 @@
...
+    cursor.execute("SELECT column_name, data_type, is_nullable FROM information_schema.columns " + \
+                   "WHERE table_schema = 'public' and table_name = '{}';".format(
+                       table_name))
+    columns = cursor.fetchall()
+
+    assert (not columns and not expected_column_tuples) \
+           or set(columns) == expected_column_tuples
+
+
+class BigCommerceStream:
```
awm33: Where's the generator for this?

target_postgres/json_schema.py
AlexanderMann: Prolly. I was doing the minimal change first. I think there's reason to do this regardless though

tests/test_sandbox.py
AlexanderMann: See our Slack convo. This just iterates over a real world stream which was snagged for testing etc.

I think this testing file should be reserved for cleansed real world data streams which we are only aiming to replay. The other testing files I think should have generators etc. which try to exercise the other various weirdness.

awm33:
@AlexanderMann Just need to address tests

PR #68, opened 2018-12-17
Housekeeping/cli tests

AlexanderMann:
# Motivation

We had a bug a while back related to cli not having testing around it. This PR adds in a few things:

- full integration testing to and from postgres
- caching for CI so that build times are faster

## Notes

This pr does not _actually_ test the data that has been loaded, nor the schema. That was viewed as scope creep. For the time being this just dumps data into the circle ci artifacts.

This will not work in the main fork till a project level env var of `TAP_GITHUB_TOKEN` has been added.
- https://github.com/singer-io/tap-github/blame/master/README.md#L34
- https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-project

## Suggested Musical Pairing

https://soundcloud.com/toroymoi/still-sound


target_postgres/__init__.py
```diff
@@ -145,7 +145,7 @@ def main(config, input_stream=None):
             host=config.get('postgres_host', 'localhost'),
             port=config.get('postgres_port', 5432),
             dbname=config.get('postgres_database'),
-            user=config.get('postgres_username'),
+            user=config.get('postgres_username', 'postgres'),
```
awm33: This defaults to the system username in unix systems. Idk if we want to override that with postgres

PR #81, opened 2019-01-10
Refactor/denest as module

AlexanderMann:
# Motivation

Start `denest`.

While working on #72 a discussion arose about longer term goals and interfaces etc. This pr seeks to split out the denesting logic so that it can be tested independently etc.

## Notes

Relies upon the work done in: https://github.com/datamill-co/target-postgres/pull/79

## Suggested Musical Pairing

https://soundcloud.com/hotchip/flutes

### DIFF
--- a/target_postgres/denest.py
+++ b/target_postgres/denest.py
@@ -2,13 +2,9 @@ from copy import deepcopy
 
 from target_postgres import json_schema
 from target_postgres.singer_stream import (
-    SINGER_BATCHED_AT,
     SINGER_LEVEL,
-    SINGER_PK,
-    SINGER_RECEIVED_AT,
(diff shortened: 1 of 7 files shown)


target_postgres/denest.py
```diff
@@ -0,0 +1,372 @@
...
+            'path': path,
+            'level': level,
+            'key_properties': keys,
+            'mappings': [],
+            'schema': {'type': 'object',
+                       'additionalProperties': False,
+                       'properties': properties}}
+
+
+def _add_singer_columns(schema, key_properties):
```
awm33: Is this only used in the module? This seems like it's the function that handles this at each level

Issue #99, opened 2019-03-11
Unknown Column Path


AlexanderMann:
# Motivation

While using `tap-hubspot`, Singer Slack User Patryk Kallinowski ran into the following error:

```
target_postgres.postgres.PostgresError: ('Exception writing records', Exception("Unknown column path: ('properties', 'num_contacted_notes', 'value') for table: ('deals',)"))
```

After uploading his `catalog.json` file generated by `tap-hubspot`, the issue appears to be around this stream/sub-schema:

```json
{
  "streams": [
    {
      "stream": "deals",
      "tap_stream_id": "deals",
      "schema": {
        "type": "object",
        "properties": {
          ...
          "properties": {
            "type": "object",
            "properties": {
              ...
              "num_contacted_notes": {
                "type": "object",
                "properties": {
                  "value": {
                    "type": ["null", "number", "string"]
                  },
                  ...
                }
              },
              ...
            }
          }
        }
      },
      "metadata": [
        ...
      ]
    }
  ]
}
```

At the very least, this error message can be improved to:

- contain more context and information for those debugging
- more descriptive/better language to identify what precisely has gone wrong
- be associated with a function which has a complete docstring

And:

- more testing to assure that whatever underlying bug (if any) is resolved

## Notes

https://singer-io.slack.com/archives/C2TGFCZEV/p1552316668073400

`catalog.json` can't be uploaded as a part of this issue, available via download in the link above.

## Suggested Musical Pairing

https://www.youtube.com/watch?v=PfrHCNo2I3M

AlexanderMann:
I think I've identified the bug here. It has to do with how we're serializing fields which have _multiple_ types. There are, I think, a couple bugs therein:

- multi type, non numeric
- single type, `[null, number]` which has values being sent as `integer`s (ie, the schema is `[null, number]`, but the values are something like `1234, 567, 8, 90` etc.

I'm fixing this bug, and _then_ I will improve the error message.

AlexanderMann:
Correction: the single type for `[null, number]` is perfectly fine.

I've found the issue. When we have a `null` value we _cannot_ assign it a "python type". ie, we don't _really_ know what value "lack of value" represents in Postgres. As such, for split column types, we don't know which column to write "nothing" to.

[...]

PR #100, opened 2019-03-11
Fix/hubspot postgres error  unknown column path

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/99

## Suggested Musical Pairing

https://soundcloud.com/mikaeladavis-music/other-lover

### DIFF
--- a/target_postgres/sql_base.py
+++ b/target_postgres/sql_base.py
@@ -588,7 +588,7 @@ class SQLInterface:
 
     def _serialize_table_record_field_name(self, remote_schema, streamed_schema, path, value_json_schema):
         """
-        Returns the appropriate remote field (column) name for `field`.
+        Returns the appropriate remote field (column) name for `path`.
 
         :param remote_schema: TABLE_SCHEMA(remote)
         :param streamed_schema: TABLE_SCHEMA(local)
@@ -616,9 +616,12 @@ class SQLInterface:
             if not mapping is None:
                 return mapping
 
-        raise Exception('Unknown column path: {} for table: {}'.format(
+        raise Exception("Column path: {} for table: {} for the value's JSONSchema {} cannot be found." \
+                        + " This means that the `remote_schema` has no compatible column for this path" \
+                        + " and value type.".format(
             path,
-            remote_schema['path']
+            remote_schema['path'],
+            simple_json_schema
         ))
 
     def serialize_table_record_null_value(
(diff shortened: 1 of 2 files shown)


AlexanderMann:
Additionally: I ran the _full_ `catalog.json` file through the sandbox testing. I did _not_ generate values for the records for each table...as that would have been...something...

Buttttt there was no issue persisting the tables to remote. So that's good!

target_postgres/sql_base.py
```diff
@@ -616,9 +616,12 @@ def _serialize_table_record_field_name(self, remote_schema, streamed_schema, pat
             if not mapping is None:
                 return mapping
 
-        raise Exception('Unknown column path: {} for table: {}'.format(
+        raise Exception("Column path: {} for table: {} for the value's JSONSchema {} cannot be found." \
+                        + " This means that the `remote_schema` has no compatible column for this path" \
```
awm33: This is probably jibberish to most users who haven't dived deep into the code. I even had to think, parse it apart, and look through the code to make sense of it.

Maybe "Column {path} for table {remote_schema['path'][0]} cannot be found in the target database"? Maybe flatten the `path` as well so it legible as a column name if possible?

target_postgres/sql_base.py
```diff
@@ -713,9 +717,19 @@ def _serialize_table_records(
                                                                      path,
                                                                      value_json_schema)
 
-                if field_name in remote_fields \
-                        and (not field_name in row
-                             or row[field_name] == NULL_DEFAULT):
+                if not field_name in row:
+                    raise Exception(
+                        "Unexpected field_name serialization! Field {} doesn't exist. Please create an issue at: {}." \
```
awm33: "Unexpected field_name serialization!" What does this mean? The column name couldn't be generated?

AlexanderMann: How about "Unexpected exception! Field {} doesn't exist. Please create..."

ie, this _shouldn't_ happen, and _is_ a bug.

target_postgres/sql_base.py
AlexanderMann: So I think the _path_ more useful here. If we get this exception, we cannot actually tell you what the name of the column is...

target_postgres/sql_base.py
AlexanderMann: Actually, I think we can just get rid of this entirely. The following check to see whether the value we've placed into the row is set will blow up with a bug worthy exception. I think we can safely ignore having a pretty exception here since this should _only_ be a bug, not something we're prepared for.

target_postgres/sql_base.py
AlexanderMann: How about `A compatible column for path {} and JSONSchema {} in table {} cannot be found.`?

awm33: I think that's better

Issue #75, opened 2019-01-03
Metrics


awm33:
The Singer standard includes metrics that are published via INFO log messages in a standard JSON format. target-postgres should publish some metrics for monitoring and diagnostic purposes.

Metrics should include:
- SQL Query timings
- Overall write_batch timings
- Batch counts
- Table persisted row counts (which may be higher than batch do to denesting)

AlexanderMann:
To note:
    - not a lot of other targets with metrics
    - less defined things/types etc.
    - timing type
    - count type
    - tagging system
    - taps for databases will have metrics in them which are most likely the metrics we want
    - `target-redshift` might want s3 information

AlexanderMann:
Spec from Singer on metrics: https://github.com/singer-io/getting-started/blob/4cc04f5b2dc2fa9173412c7cd84df4bf2ed0ca91/docs/SYNC_MODE.md#metric-messages

AlexanderMann:
Code in [`singer-python`](https://github.com/singer-io/singer-python/blob/a70e03f108307ee29f02811303bc95324650f10b/singer/metrics.py)

AlexanderMann:
Timer examples:
- https://github.com/singer-io/tap-mysql/blob/5bac530b6f5fe07194311cbb5626cc3f87be7e3b/tap_mysql/__init__.py#L639-L652

Counter examples:
- https://github.com/singer-io/tap-mysql/blob/cfd04fcf8cf60d5c679207abaa42d11af0c996be/tap_mysql/sync_strategies/common.py#L173-L175
- https://github.com/singer-io/tap-postgres/blob/3dce12abc2729fee2719d782739b0945591ec0c9/tap_postgres/sync_strategies/incremental.py#L56
- https://github.com/singer-io/tap-postgres/blob/3dce12abc2729fee2719d782739b0945591ec0c9/tap_postgres/sync_strategies/full_table.py#L41
- https://github.com/singer-io/tap-postgres/blob/3dce12abc2729fee2719d782739b0945591ec0c9/tap_postgres/sync_strategies/full_table.py#L98

AlexanderMann:
Soooo... `tap-postgres` doesn't do anything with tags (confirmed in this repo's usage of it as well, ie CCI logs). `tap-mysql` _however_ does, but not using the "recommended" `singer-python` mechanisms:

```
    with metrics.record_counter(None) as counter:
        counter.tags['database'] = database_name
[...]

PR #108, opened 2019-04-01
Feature: init singer.metrics usage

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/75

## Notes

Whereas the `tap`s for SQL _know_ which table they are persisting to, our `stream`s are potentially nested. For this reason, we include the following tags:

- `path`: `tuple` of `string`s representing the denested path to a given table
- `table`: optional `string` representing the remote table we are dealing with
- `schema`: `string` representing the remote schema we're dealing with
- `database`: `string` representing the remote database we're dealing with

## Testing

Output provided by the CCI tests:

```
INFO METRIC: {"type": "timer", "metric": "job_duration", "value": 0.013872385025024414, "tags": {"job_type": "upsert_table_schema", "path": ["stargazers"], "database": "target_postgres_test", "schema": "public", "table": "stargazers", "status": "succeeded"}}
INFO Writing table batch with 18 rows for `('stargazers',)`...
INFO METRIC: {"type": "timer", "metric": "job_duration", "value": 0.011026144027709961, "tags": {"job_type": "upsert_table_schema", "path": ["tmp_c85777c0_12cd_41a0_ad96_089ea9418713"], "database": "target_postgres_test", "schema": "public", "table": "tmp_c85777c0_12cd_41a0_ad96_089ea9418713", "status": "succeeded"}}
INFO METRIC: {"type": "counter", "metric": "record_count", "value": 18, "tags": {"count_type": "table_rows_persisted", "path": ["stargazers"], "database": "target_postgres_test", "schema": "public", "table": "stargazers"}}
INFO METRIC: {"type": "timer", "metric": "job_duration", "value": 0.03966832160949707, "tags": {"job_type": "table", "path": ["stargazers"], "database": "target_postgres_test", "schema": "public", "table": "stargazers", "status": "succeeded"}}
INFO METRIC: {"type": "counter", "metric": "record_count", "value": 18, "tags": {"count_type": "batch_rows_persisted", "path": ["stargazers"], "database": "target_postgres_test", "schema": "public"}}
INFO METRIC: {"type": "timer", "metric": "job_duration", "value": 0.041277408599853516, "tags": {"job_type": "batch", "path": ["stargazers"], "database": "target_postgres_test", "schema": "public", "status": "succeeded"}}
...
```

## Suggested Musical Pairing
https://soundcloud.com/10cc-official/art-for-arts-sake-album?in=drunken-ambassador/sets/interval

### DIFF
--- a/target_postgres/postgres.py
+++ b/target_postgres/postgres.py
@@ -114,6 +114,10 @@ class PostgresTarget(SQLInterface):
         if self.persist_empty_tables:
             self.LOGGER.debug('PostgresTarget is persisting empty tables')
 
+    def metrics_tags(self):
+        return {'database': self.conn.get_dsn_parameters().get('dbname', None),
+                'schema': self.postgres_schema}
+
     def write_batch(self, stream_buffer):
         if not self.persist_empty_tables and stream_buffer.count == 0:
             return None
--- a/target_postgres/sql_base.py
+++ b/target_postgres/sql_base.py
@@ -53,8 +53,16 @@ class SQLInterface:
     IDENTIFIER_FIELD_LENGTH = NotImplementedError('`IDENTIFIER_FIELD_LENGTH` not implemented.')
     LOGGER = singer.get_logger()
 
-    def _set_metrics_tags(self, metric, job_type, path):
+    def _set_timer_tags(self, metric, job_type, path):
         metric.tags['job_type'] = job_type
+        metric.tags['path'] = path
+
+        metric.tags.update(self.metrics_tags())
+
+        return metric
+
+    def _set_counter_tags(self, metric, counter_type, path):
+        metric.tags['count_type'] = counter_type
         metric.tags['path'] = path
 
         metric.tags.update(self.metrics_tags())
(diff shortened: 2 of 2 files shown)


awm33:
@AlexanderMann Why can't we know the table name for row counts and have to use paths? Even nested arrays eventually become tables. We place them in a table when we merge the tmp table or the current batch size for a full-table rep.

AlexanderMann:
@awm33 I put in place both the path and the table:

```
{"path": ["pull_requests"], "database": "target_postgres_test", "schema": "public", "table": "pull_requests"}
```

For the top level stream we can't (necessarily) have a table name since that's cumulative inserts from all of the various nested objects. We could put in place a `stream` tag as well? Personally I think _more_ tags than fewer are useful for slicing metrics up, but that's just my experience building observability and monitoring systems.

AlexanderMann:
@awm33 I redact my previous comment. The `stream` name is a little bit harder to get in place for this. We don't pass that information down into `SQLBase` presently (I think that's actually one of the refactor things we had wanted to do) and to get it down in there would be a bit of work.

For the time being, I'd vote that we _don't_ include the `stream` tag, but if we decide that it's actually more useful to have (etc.) I can bite that off as a part of this work.

awm33:
@AlexanderMann Ok that may make sense then. I think what we'll probably want, which are aggregates, are for a given run, or over period of time, row counts per table in the data warehouse. It looks like the reverse is possible too given the path?

AlexanderMann:
@awm33 back to you!

awm33:
@AlexanderMann So beautiful 😍 . Imma query the crap out of these metrics once we start running this in prod.

Issue #109, opened 2019-04-04
No data in database if nested objects within array


CStejmar:
Hi,
I think I have encountered a problem with nested objects in arrays. An object in an array is fine but when I have an object within another object in an array (array->object->object) problem arises. No data is seen in that last level object even if I can see in the record data that values exist. If I instead have an array at the last level (array->object->array) it behaves as expected.

I have done tests with this in the module `test_sandbox` using a modified `CATS_SCHEMA` and the `FakeStream.generate_record_message`(also modified to match my schema) to generate test data. If I in the test make the last object's properties optional by `"type": ["null", "string"]` for example, everything passes because the object's properties are all null. However, if I use `"type": "string"` for a property, the test fails with:<br/>
`CRITICAL ('Exception writing records', IntegrityError('null value in column "adoption__immunizations__vaccination_type__shot" violates not-null constraint\nDETAIL:  Failing row contains (Rabies, 2537-09-12 15:34:00+02, null, 1, 1554384634, 0).\nCONTEXT:  COPY tmp_36a9e5fb_ac07_49f0_ac26_8fa9f1f69885, line 1: "Rabies,2537-09-12 13:34:00.0000+00:00,NULL,1,1554384634,0"\n'))`<br/>

To me it seems like this case isn't handled and I can only find tests of `array_of_array` and `object_of_object` in the tests directory. Does anyone have a clue and can point me in the right direction?<br/>

I will attach a snippet of what I added to the  `test_sandbox.py` file so that you can test my examples and easier understand my problem.
<br/>
```python
class CatStreamObject(SandboxStream):
    stream = [
        {"type": "SCHEMA",
         "stream": "cats",
         "schema": {
             "additionalProperties": False,
             "properties": {
                 "id": {"type": "integer"},
                 "name": {"type": ["string"]},
                 "paw_size": {"type": ["integer"],
                              "default": 314159},
                 "paw_colour": {"type": "string", "default": ""},
                 "flea_check_complete": {"type": ["boolean"],
                                         "default": False},
                 "pattern": {"type": ["null", "string"]},
                 "age": {"type": ["null", "integer"]},
                 "adoption": {"type": ["object", "null"],
                              "properties": {"adopted_on": {
                                  "type": ["null", "string"],
                                  "format": "date-time"},
                                  "was_foster": {
                                      "type": "boolean"},
                                  "immunizations": {
                                      "type": ["null",
                                               "array"],
                                      "items": {
                                          "type": [
[...]

PR #110, opened 2019-04-04
FIX: Nested Records of objects of objects.

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/109

This PR exists to lay down test cases which our denesting _should_ handle but currently doesn't.

## Suggested Musical Pairing

Springtime in Milwaukee.

### DIFF
--- a/tests/test_denest.py
+++ b/tests/test_denest.py
@@ -199,29 +199,40 @@ def test__records__nested():
                                   "c": {
                                       "type": "object",
                                       "properties": {
-                                          "d": {
-                                              "type": "integer"}}}}}}}}}},
+                                          "d": {"type": "integer"},
+                                          "e": {"type": "array",
+                                                "items": {"type": "object",
+                                                          "properties": {
+                                                              "f": {"type": "string"},
+                                                              "g": {"type": "boolean"}}}}}}}}}}}}},
         [],
         [{"a": {"b": []}},
-         {"a": {"b": [{"c": 1}]}},
-         {"a": {"b": [{"c": 12},
-                      {"c": 123}]}},
-         {"a": {"b": [{"c": 1234},
-                      {"c": 12345},
-                      {"c": 123456}]}}])
+         {"a": {"b": [{"c": {"d": 1}}]}},
+         {"a": {"b": [{"c": {"d": 12}},
+                      {"c": {"d": 123}}]}},
+         {"a": {"b": [{"c": {"d": 1234}},
+                      {"c": {"d": 12345}},
+                      {"c": {"d": 123456}}]}},
+         {"a": {"b": [{"c": {"e": [{"f": "hello",
+                                    "g": True},
+                                   {"f": "goodbye",
+                                    "g": True}]}}]}}])
 
     print('denested:', denested)
 
-    assert 2 == len(denested)
+    assert 3 == len(denested)
 
     root_table_checked = False
     nested_table_checked = False
+    nested_nested_table_checked = False
 
     for table_batch in denested:
         if tuple() == table_batch['streamed_schema']['path']:
             root_table_checked = True
 
-            assert 4 == len(table_batch['records'])
+            assert {} == table_batch['streamed_schema']['schema']['properties']
+
+            assert 5 == len(table_batch['records'])
 
             for record in table_batch['records']:
                 assert {} == record
@@ -230,14 +241,30 @@ def test__records__nested():
         if ('a', 'b') == table_batch['streamed_schema']['path']:
             nested_table_checked = True
 
-            assert 6 == len(table_batch['records'])
+            assert {'type': 'integer'} == table_batch['streamed_schema']['schema']['properties'][('c', 'd')]
+
+            assert 7 == len(table_batch['records'])
 
             for record in table_batch['records']:
-                assert 'integer' == record[('c',)][0]
-                assert int == type(record[('c',)][1])
+                assert 'integer' == record[('c', 'd')][0]
+                assert int == type(record[('c', 'd')][1])
+
+    for table_batch in denested:
+        if ('a', 'b', 'c', 'e') == table_batch['streamed_schema']['path']:
+            nested_nested_table_checked = True
+
+            assert {'type': 'string'} == table_batch['streamed_schema']['schema']['properties'][('f',)]
+            assert {'type': 'boolean'} == table_batch['streamed_schema']['schema']['properties'][('g',)]
+
+            assert 2 == len(table_batch['records'])
+
+            for record in table_batch['records']:
+                assert 'string' == record[('f',)][1]
+                assert record[('g',)][1]
 
     assert root_table_checked
     assert nested_table_checked
+    assert nested_nested_table_checked
 
     for table_batch in denested:
         assert [] == errors(table_batch)
--- a/tests/test_denest.py
+++ b/tests/test_denest.py
@@ -230,7 +230,7 @@ def test__records__nested__tables():
 def _get_table_batch_with_path(table_batches, path):
     for table_batch in table_batches:
         if path == table_batch['streamed_schema']['path']:
-            return path
+            return table_batch
     raise Exception('Could not find table_batch with path: {}'.format(path))
 
 
--- a/tests/test_denest.py
+++ b/tests/test_denest.py
@@ -184,3 +184,60 @@ def test__schema__nested_arrays_add_tables():
 
         print('PASSED')
         print()
+
+
+def test__records__nested():
+    denested = denest.to_table_batches({
+        "properties": {
+            "a": {"type": "object",
+                  "properties": {
+                      "b": {
+                          "type": "array",
+                          "items": {
+                              "type": "object",
+                              "properties": {
+                                  "c": {
+                                      "type": "object",
+                                      "properties": {
+                                          "d": {
+                                              "type": "integer"}}}}}}}}}},
+        [],
+        [{"a": {"b": []}},
+         {"a": {"b": [{"c": 1}]}},
+         {"a": {"b": [{"c": 12},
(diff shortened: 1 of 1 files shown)


tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+                                   {"f": "goodbye",
+                                    "g": True}]}}]}}])
+
+    print('denested:', denested)
+
+    assert 3 == len(denested)
+
+    root_table_checked = False
+    nested_table_checked = False
+    nested_nested_table_checked = False
```
CStejmar: Should we split up the test in three parts like this in the same test? We could place these three in separate test functions or just have one loop through `denested` in one test function in my opinion. What do you think is the way to go here?

tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+
+    for table_batch in denested:
+        if tuple() == table_batch['streamed_schema']['path']:
+            root_table_checked = True
+
+            assert {} == table_batch['streamed_schema']['schema']['properties']
+
+            assert 5 == len(table_batch['records'])
```
CStejmar: This is not really related to this PR:
When I was running the tests and looking at the printed denested data I started thinking about this line. Why is there empty dicts ({}) added to the records field in the `_denest_record` function? Even if the record doesn't contain any items we still add an empty dict. Is that desired?

We could change this with a simple check in `_denest_record` when appending denested records:
```python
if denested_record != {}:
        records_map[table_path].append(denested_record)
```
If we want to change this we should do it in another PR. Also, this change makes some length asserts fail and that needs to be addressed as well in that case.

tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+
+            for record in table_batch['records']:
+                assert {} == record
+
+    for table_batch in denested:
+        if ('a', 'b') == table_batch['streamed_schema']['path']:
+            nested_table_checked = True
+
+            assert {'type': 'integer'} == table_batch['streamed_schema']['schema']['properties'][('c', 'd')]
```
CStejmar: Needs to be `assert {'type': ['integer']} == ...` to match result from `table_batch`

tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+
+            assert 7 == len(table_batch['records'])
+
+            for record in table_batch['records']:
+                assert 'integer' == record[('c', 'd')][0]
```
CStejmar: Above this line I needed to add:
```python
if record == {}:
    continue
```
due to one empty dict. Accessing key `('c', 'd')` will otherwise fail.

tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+                assert int == type(record[('c', 'd')][1])
+
+    for table_batch in denested:
+        if ('a', 'b', 'c', 'e') == table_batch['streamed_schema']['path']:
+            nested_nested_table_checked = True
+
+            assert {'type': 'string'} == table_batch['streamed_schema']['schema']['properties'][('f',)]
```
CStejmar: 'string' -> ['string']

CStejmar: Same goes for line under but instead: 'boolean' -> ['boolean']

tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+            assert {'type': 'boolean'} == table_batch['streamed_schema']['schema']['properties'][('g',)]
+
+            assert 2 == len(table_batch['records'])
+
+            for record in table_batch['records']:
+                assert 'string' == record[('f',)][1]
+                assert record[('g',)][1]
```
CStejmar: I would change this to:
```python
for record in table_batch['records']:
    assert 'string' == record[('f',)][0]
    assert str == type(record[('f',)][1])

    assert 'boolean' == record[('g',)][0]
    assert bool == type(record[('g',)][1])
```

CStejmar: Just to have it consistent with what we have in `nested_table_checked` loop:
```python
assert 'integer' == record[('c', 'd')][0]
assert int == type(record[('c', 'd')][1])
```

tests/test_denest.py
```diff
@@ -184,3 +184,87 @@ def test__schema__nested_arrays_add_tables():
...
+
+    assert root_table_checked
+    assert nested_table_checked
+    assert nested_nested_table_checked
+
+    for table_batch in denested:
+        assert [] == errors(table_batch)
```
CStejmar: At the end maybe we should add:
```python
print('PASSED')
print()
```
I have seen this in other test functions so to keep the consistency we should add that as well.

tests/test_denest.py
AlexanderMann: I like that idea. I'll refactor for that. My main concern originally with going that route, was that there'll be a modicum of "helper" code around these tests which will be obfuscating the actual test. Additionally, I was worried that previously we did _really_ precise tests which looked for specific things and hence had specific schemas. The bug we're experiencing is because we didn't have a gnarlier schema and records to pick apart etc. As such, I'm going to plan on the three tests sharing the same schema and records.

That all sound right to you?

tests/test_denest.py
AlexanderMann: Yeah, it's necessary. So if you get rid of that, then there's no way for our code to create a key for the denested values to relate _back_ to.

ie, think of if you had a schema like this:

```json
{'properties':
  {'a': {'items': [...]},
   'b': {'items': [...]}}}
```

So `a` and `b` _relate_ to each other in that they're on the same parent right? But when we denest, this'll turn into:

```
root: {}
root__a: {...}
root__b: {...}
```

Since we generate uuids for tables which don't have `key_properties` set, we can then relate the child tables and records back to each other in postgres.

If we get rid of the record, it indicates to everything else that there's just nothing there...

(now, mind you, the code _above_ here where we're setting up the `Singer Stream` etc., actually injects a `key_properties` and uuids etc. so you can't ever have this case of empty tables. There'll always be the StitchData `_sdc` columns/fields/properties)

tests/test_denest.py
AlexanderMann: Good call

AlexanderMann:
@CStejmar 🏓

tests/test_denest.py
CStejmar: Sounds perfect!

tests/test_denest.py
CStejmar: Ah I see, thanks for the explanation! Makes perfect sense 👍

tests/test_denest.py
CStejmar: Ok great! Then we remove it :)

tests/test_denest.py
```diff
@@ -182,89 +179,104 @@ def test__schema__nested_arrays_add_tables():
...
-         {"a": {"b": [{"c": {"d": 12}},
-                      {"c": {"d": 123}}]}},
-         {"a": {"b": [{"c": {"d": 1234}},
-                      {"c": {"d": 12345}},
-                      {"c": {"d": 123456}}]}},
-         {"a": {"b": [{"c": {"e": [{"f": "hello",
-                                    "g": True},
-                                   {"f": "goodbye",
-                                    "g": True}]}}]}}])
+NESTED_SCHEMA = {
```
CStejmar: 👍

tests/test_denest.py
```diff
@@ -182,89 +179,104 @@ def test__schema__nested_arrays_add_tables():
...
+                                  "type": "object",
+                                  "properties": {
+                                      "d": {"type": "integer"},
+                                      "e": {"type": "array",
+                                            "items": {"type": "object",
+                                                      "properties": {
+                                                          "f": {"type": "string"},
+                                                          "g": {"type": "boolean"}}}}}}}}}}}}}
+
+NESTED_RECORDS = [{"a": {"b": []}},
```
CStejmar: 👍

tests/test_denest.py
```diff
@@ -182,89 +179,104 @@ def test__schema__nested_arrays_add_tables():
...
+                               {"c": {"d": 12345}},
+                               {"c": {"d": 123456}}]}},
+                  {"a": {"b": [{"c": {"e": [{"f": "hello",
+                                             "g": True},
+                                            {"f": "goodbye",
+                                             "g": True}]}}]}}]
+
+
+def test__records__nested__tables():
+    denested = denest.to_table_batches(NESTED_SCHEMA, [], NESTED_RECORDS)
```
CStejmar: 👍

tests/test_denest.py
```diff
@@ -182,89 +179,104 @@ def test__schema__nested_arrays_add_tables():
...
+                ('a', 'b', 'c', 'e')}
+        assert [] == errors(table_batch)
 
-            assert {} == table_batch['streamed_schema']['schema']['properties']
 
-            assert 5 == len(table_batch['records'])
+def _get_table_batch_with_path(table_batches, path):
+    for table_batch in table_batches:
+        if path == table_batch['streamed_schema']['path']:
+            return path
```
CStejmar: Here we need to return `table_batch` instead of `path` otherwise we fail the tests on trying to access 'schema' etc. I know you meant do that since the function name is `_get_table_batch_with_path` :).

AlexanderMann: 🤦‍♂️ this has been a hilariously tough thing for me to do as I don't actually _have_ the working code etc.

CStejmar: Yeah I can imagine that 🙂. Anyway, great work! 🚀

PR #113, opened 2019-04-16
CHANGELOG: Bug for 0.1.7

AlexanderMann:
# Motivation

Prep for 0.1.7

## Suggested Musical Pairing

Dogs chewing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.1.7
+
+- **BUG FIX:** A bug was identified for de-nesting.
+  - [ISSUE LINK](https://github.com/datamill-co/target-postgres/issues/109)
+  - [FAILING TESTS LINK](https://github.com/datamill-co/target-postgres/pull/110)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## 0.1.7
+
+- **BUG FIX:** A bug was identified for de-nesting.
+  - [ISSUE LINK](https://github.com/datamill-co/target-postgres/issues/109)
+  - [FAILING TESTS LINK](https://github.com/datamill-co/target-postgres/pull/110)
+  - [FIX LINK](https://github.com/datamill-co/target-postgres/pull/111)
+  - Subtables with subtables do not serialize column names correctly
```
CStejmar: "do not" -> "did not" ?

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
+    - The column names ended up having the _table names_ (paths) prepended on them
+
```
CStejmar: Here you could add something like this (since it was also part of the problem):
`- Table schemas and records did not match which led to data not being ingested into database`

PR #120, opened 2019-05-31
Send STATE records to stdout

### LINKED ISSUE #117: Pass STATE messages to stdout
In order for taps that allow incremental processing to work properly, they must have state data that informs where to pick up during the next sync. Currently, this target does not emit state messages.

This behavior seems to run counter to the specs. Notably, the [Running and Developing Singer Taps and Targets](https://github.com/singer-io/getting-started/blob/master/docs/RUNNING_AND_DEVELOPING.md#guidelines) document states:

> Write State messages to stdout once all data that appeared in the stream before the State message has been processed by the Target. Note that although the State message is sent into the target, in most cases the Target's process won't actually store it anywhere or do anything with it other than repeat it back to stdout.

bradleyprice:
Fixes #117 

This will pass all `STATE` records to stdout.

The command could now be:

`tap | target >> state.json`

And the resulting state file would look something like this:

```
{ "users": 1 }
{ "users": 2 }
```

With this approach, if the end user only cares about the latest
`STATE` record, they can manually handle updating their state file:

`tail -1 state.json > state.json.tmp && mv state.json.tmp state.json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ pip install singer-target-postgres
    ```bash
    ~/.virtualenvs/tap-something/bin/tap-something \
      | ~/.virtualenvs/target-postgres/bin/target-postgres \
-       --config ~/singer.io/target_postgres_config.json
+       --config ~/singer.io/target_postgres_config.json >> state.json
    ```
 
 ### Config.json
@@ -93,7 +93,6 @@ _The above is copied from the [current list of versions](https://www.postgresql.
 
 ## Known Limitations
 
-- Ignores `STATE` Singer messages.
 - Requires a [JSON Schema](https://json-schema.org/) for every stream.
 - Only string, string with date-time format, integer, number, boolean,
   object, and array types with or without null are supported. Arrays can
--- a/target_postgres/target_tools.py
+++ b/target_postgres/target_tools.py
@@ -36,6 +36,7 @@ def stream_to_target(stream, target, config={}):
     :return: None
     """
 
+    state_writer = sys.stdout
     streams = {}
     try:
         if not config.get('disable_collection', False):
@@ -55,7 +56,8 @@ def stream_to_target(stream, target, config={}):
                           invalid_records_threshold,
                           max_batch_rows,
                           max_batch_size,
-                          line)
+                          line,
+                          state_writer)
             if line_count > 0 and line_count % batch_detection_threshold == 0:
                 _flush_streams(streams, target)
             line_count += 1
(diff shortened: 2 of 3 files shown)


micaelbergeron:
@bradleyprice I just want to bring something to your attention here. Albeit this implementation "works" I think there will be a problem if the target fails with records buffered.

The way the STATE message should be handled in the target is that it should make sure all records are persisted **before** emitting the STATE message back.

There are multiple ways to handle that, from simply flushing the whole buffers when a STATE message is received, but that might have some performance issues.

Another approach is to introspect the STATE messages and flush only the corresponding records, but that is much more work.

Another approach would be to ignore STATE messages until N-records are buffered.

I hope this is helpful to you.

bradleyprice:
@micaelbergeron Thanks for the feedback! I'll try to work on the implementation and see if I can make it handle these failure points.

awm33:
@bradleyprice Thank for contributing!

@bradleyprice @micaelbergeron Yeahh, the difficulty here is that we batch by stream and records could be mixed. A potential solution, though complex:
- Use https://github.com/singer-io/singer-python/blob/master/singer/statediff.py to only track state messages that are different from the previous. (Stitch does this internally when running taps)
- Create a structure to track unique states to stream buffers that contain them and only flush state when all buffers associated with a state have been flushed. EG have a set with each buffer per unique state, you could hook into the buffer's flush function to remove itself from the set, when the set is empty, flush state.

Or just flush it at the end, maybe only the last state. Singer tap streams should be idempotent, so if the state/bookmarks are off, it should just mean a longer load time next run, not data discrepancies. Even that may be an improvement for those seeking this functionlality.

1) I'd like this to be optional defaulting to false, since it doesn't do this currently. This can be done using the config.
2) Can we use the full message structure? eg {"type":"STATE", "value":...}. I realize target-stitch doesn't do this, but it feels like communications between Singer components should use the message spec. All of state handling isn't well spec'ed out in Singer, and the most common use case laid out is only using it against Stitch.

bradleyprice:
@awm33 I'll take a look into how stitch is handling state and see if I can come up with a better solution. 

Thanks for the feedback!

robmoore:
@bradleyprice It might be worth reviewing the work that @micaelbergeron did implementing a [Snowflake target] (https://gitlab.com/meltano/target-snowflake/blob/master/target_snowflake/target_snowflake.py). In particular, I think that `StateBuffer` on line 114 might be helpful.

bradleyprice:
@robmoore Thanks! I'll take a look at that as soon as I have a chance.

AlexanderMann:
Going to vote we close this in favour of #130.

PR #121, opened 2019-05-31
Bump psycopg2 from 2.7.7 -> 2.8.2

### LINKED ISSUE #119: PsycoPG2: Cannot bump to latest version
# Motivation

While releasing `0.1.8`, I tried to put PsycoPG2 to its latest dependency and ran into significant issues. I am currently _unsure_ as to what happened.

[Failing CI build here](https://circleci.com/workflow-run/3482254f-c784-4fa3-8cdd-91ac6c1f335f)

Investigation into _why_ this is happening (even just noting what in the CHANGELOG for PyscoPG2 has changed which _might relate_) is welcomed!

bradleyprice:
Fixes #119

psycopg2 2.8 release introduced mappings between Postgresql errors
and python exceptions.

I assume this came with additional, more granular exceptions that are
now subclassed.

If the original code had been throwing the `IntegrityError` exception, the
code would have still worked, since the `NotNullValidation` is a subclass.
However, since the exception is wrapped in `postgres.PostgresError` and
it's using a matcher, the matching string had to be updated to match the
new exception being thrown.

REFERENCES:
* http://initd.org/psycopg/articles/2019/04/04/psycopg-28-released/
* http://initd.org/psycopg/docs/errors.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,8 +20,8 @@ setup(
     py_modules=['target_postgres'],
     install_requires=[
         'arrow==0.13.0',
-        'psycopg2==2.7.7',
-        'psycopg2-binary==2.7.7',
+        'psycopg2==2.8.2',
+        'psycopg2-binary==2.8.2',
         'singer-python==5.6.0'
     ],
     setup_requires=[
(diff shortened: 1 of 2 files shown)


AlexanderMann:
@bradleyprice this looks good to me. I'm wondering if this'll end up being a problem with Redshift, but I do want to keep up to date with PsycoPG2.

PR #122, opened 2019-05-31
Fix typo disttribution -> distribution

bradleyprice:


awm33:
@bradleyprice Thanks for catching that!!

PR #124, opened 2019-06-08
Enable Configuration of SSL

citruspi:
Enables configuration of SSL when connecting to Postgres with the following (optional) config keys:

- `postgres_sslmode`
- `postgres_sslcert`
- `postgres_sslkey`
- `postgres_sslrootcert`
- `postgres_sslcrl`


awm33:
@citruspi Thank you for contributing!

This looks pretty straight forward, I'm going to merge. It will be in our next release, which isn't scheduled, but I think will be once we get some dependency upgrade PRs merged and potentially STATE management.

PR #126, opened 2019-06-19
windows equivalent to run target

aroder:


awm33:
@aroder Thanks for the contribution!

PR #130, opened 2019-07-24
Performant STATE message flush

airhorns:
This is on top of #120 and a follow up to #128. 

In order for orchestrating systems to properly keep track of the STATE of Singer streams, target-postgres needs to emit the STATE messages it receives from the tap to stdout for the orchestrator to persist. This keeps target-postgres Singer-spec compliant and lets users pass that STATE back into the tap to incrementally load from any big sources.

The Singer spec however does not specify what exactly is in STATE messages, so, they can relate to any or all of the active streams. That means that target-postgres doesn't and can't know which records that it may have buffered in memory are "covered" by an incoming STATE record. If target-postgres eagerly emitted the STATE record to stdout, but didn't flush all the records, the system becomes open to inconsistency. If the STATE message were persisted by the outside orchestrator, but then the process crashed later, the buffers in target-postgres process would be lost and data dropped. The implementation from #120 suffers from this bug, see #120 for more discussion.

This commit prevents this bug by "late" flushing buffered STATE messages when all the records that arrived prior to that STATE message have been flushed. STATE messages are "delayed" until the records that came before have all been put on a buffered stream and then flushed. target-snowflake implements this here: https://gitlab.com/meltano/target-snowflake/blob/master/target_snowflake/target_snowflake.py.

This is accomplished by keeping a low watermark of the least recently arriving record for each stream, and upon flushing a buffer, checking to see if there are any unflushed STATE messages that have become safe to flush because they are below that watermark. The way to think about this is as a "safety cursor" that lags behind the incoming stream, pointing to the point in the stream where all records before that point have been saved to the database. Some records ahead of that point will likely have been saved, as different streams records may be interleaved or arriving at different rates, but because of the whole "states are on one timeline different than each stream" thing, the STATE message has to wait for all prior records to be flushed.

This is implemented using a new StateTracker class, which wraps a bit of business logic around the old streams dict that is used to hold the BufferedSingerStream objects. I'm happy to get feedback on the style or further refactorings that this might prompt so please review away. I tried to make this change in a not super invasive way such that we could potentially chunk up further changes into future PRs.

### DIFF
--- a/target_postgres/state_tracker.py
+++ b/target_postgres/state_tracker.py
@@ -1,0 +1,24 @@
+import json
+import sys
+
+class StateTracker:
+  def __init__(self, target, emit_states):
+    self.target = target
+    self.emit_states = emit_states
+
+    self.streams = {}
+    self.last_emitted_state = None
+    self.state_queue = []
+
+  def flush_streams(self, force=False):
+    for stream_buffer in self.streams.values():
+        if force or stream_buffer.buffer_full:
+          self.target.write_batch(stream_buffer)
+          stream_buffer.flush_buffer()
+
+
+  def emit_state(self, value):
+    if self.emit_states:
+      line = json.dumps(value)
+      sys.stdout.write("{}\n".format(line))
+      sys.stdout.flush()
--- a/target_postgres/state_tracker.py
+++ b/target_postgres/state_tracker.py
@@ -1,24 +1,73 @@
+from collections import deque
 import json
 import sys
 
+
+class TargetError(Exception):
+    """
+    Raise when there is an Exception streaming data to the target.
+    """
+
+
 class StateTracker:
-  def __init__(self, target, emit_states):
-    self.target = target
-    self.emit_states = emit_states
+    """
+    Object to track STATE messages coming in from the tap and which streams need to be flushed before they can be safely emitted.
+    Because Singer taps don't have a standard way of expressing which streams correspond to which STATEs, the target can only safely
+    emit a STATE message once all the records that came in prior to that STATE in the stream. Because target-postgres buffers
+    the records in BufferedSingerStreams, the STATE messages need to be delayed until all the records that came before them have been
+    saved to the database from their buffers.
+    """
 
-    self.streams = {}
-    self.last_emitted_state = None
-    self.state_queue = []
+    def __init__(self, target, emit_states):
+        self.target = target
+        self.emit_states = emit_states
 
-  def flush_streams(self, force=False):
-    for stream_buffer in self.streams.values():
-        if force or stream_buffer.buffer_full:
-          self.target.write_batch(stream_buffer)
-          stream_buffer.flush_buffer()
+        self.streams = {}
+        self.stream_add_watermarks = {}
+        self.stream_flush_watermarks = {}
 
+        self.state_queue = deque()  # contains dicts of {'state': <state blob>, 'watermark': number}
+        self.message_counter = 0
 
-  def emit_state(self, value):
-    if self.emit_states:
-      line = json.dumps(value)
-      sys.stdout.write("{}\n".format(line))
-      sys.stdout.flush()
+    def register_stream(self, stream, buffered_stream):
+        self.streams[stream] = buffered_stream
+        self.stream_add_watermarks[stream] = 0
+        self.stream_flush_watermarks[stream] = 0
+
+    def flush_streams(self, force=False):
+        for (stream, stream_buffer) in self.streams.items():
+            if force or stream_buffer.buffer_full:
+                self.target.write_batch(stream_buffer)
+                stream_buffer.flush_buffer()
+                self.stream_flush_watermarks[stream] = self.stream_add_watermarks[stream]
+
+        self._emit_safe_queued_states(force=force)
+
+    def handle_state_message(self, value):
+        if self.emit_states:
+            self.state_queue.append({'state': value, 'watermark': self.message_counter})
+            self._emit_safe_queued_states()
+
+    def handle_record_message(self, stream, line_data):
+        if stream not in self.streams:
+            raise TargetError('A record for stream {} was encountered before a corresponding schema'.format(stream))
+
+        self.message_counter += 1
+        self.stream_add_watermarks[stream] = self.message_counter
+        self.streams[stream].add_record_message(line_data)
+
+    def _emit_safe_queued_states(self, force=False):
+        # State messages that occured before the least recently flushed record are safe to emit.
+        # If they occurred after some records that haven't yet been flushed, they aren't safe to emit.
+        # Because records arrive at different rates from different streams, we take the earliest unflushed record as the threshold for what
+        # STATE messages are safe to emit.
+        all_flushed_watermark = min(self.stream_flush_watermarks.values(), default=0)
+        emittable_state = None
+
+        while len(self.state_queue) > 0 and (force or self.state_queue[0]['watermark'] <= all_flushed_watermark):
+            emittable_state = self.state_queue.popleft()['state']
+
+        if emittable_state:
+            line = json.dumps(emittable_state)
+            sys.stdout.write("{}\n".format(line))
+            sys.stdout.flush()
--- a/target_postgres/state_tracker.py
+++ b/target_postgres/state_tracker.py
@@ -1,6 +1,7 @@
 from collections import deque
 import json
 import sys
+import singer.statediff as statediff
 
 
 class TargetError(Exception):
@@ -28,6 +29,7 @@ class StateTracker:
 
         self.state_queue = deque()  # contains dicts of {'state': <state blob>, 'watermark': number}
         self.message_counter = 0
+        self.last_emitted_state = None
 
     def register_stream(self, stream, buffered_stream):
         self.streams[stream] = buffered_stream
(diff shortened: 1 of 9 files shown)


awm33:
@airhorns This is great!

I added a few comments inline.

But once thing to think about, since `StateTracker`  is now responsible for handling certain messages as well as the root for streams, perhaps it should really be called `StreamTracker` or `Streams` or `MessageHandler`?

airhorns:
@awm33 ok, I will rename it. Also I don't see your comments, is your review maybe pending still? Also, also, do you want me to go further with the refactor right now and move more into the `StreamTracker` or is how it is suitable for you?

AlexanderMann:
Hey @airhorns, I'm dedicating my day to all things Target-Postgres and am hoping to get you critical feedback here etc.

awm33:
@airhorns @AlexanderMann Sorry! Forgot to hit Submit Review

target_postgres/state_tracker.py
```diff
@@ -0,0 +1,78 @@
...
+        self.message_counter += 1
+        self.stream_add_watermarks[stream] = self.message_counter
+        self.streams[stream].add_record_message(line_data)
+
+    def _emit_safe_queued_states(self, force=False):
+        # State messages that occured before the least recently flushed record are safe to emit.
+        # If they occurred after some records that haven't yet been flushed, they aren't safe to emit.
+        # Because records arrive at different rates from different streams, we take the earliest unflushed record as the threshold for what
+        # STATE messages are safe to emit.
+        all_flushed_watermark = min(self.stream_flush_watermarks.values(), default=0)
```
AlexanderMann: If we wanted to get crazy about performance, this'd prolly make sense to be a min heap. I'd personally vote for this to be tackled later if there're tonnes of open streams at any given time etc.

tests/test_target_tools.py
```diff
@@ -70,3 +76,174 @@ def test_loading__invalid__records__threshold():
...
+
+    target_tools.stream_to_target(test_stream(), target, config=config)
+
+    # The final state message should have been outputted after the last dog records were loaded despite not reaching one full flushable batch
+    output = filtered_output(capsys)
+    assert len(output) == 1
+    assert json.loads(output[0])['test'] == 'state-4'
+
+
+def test_state__doesnt__emit__when__it__isnt__different__than__the__previous__emission(capsys):
```
AlexanderMann: nit: underscores 👍

tests/test_target_tools.py
```diff
@@ -70,3 +76,174 @@ def test_loading__invalid__records__threshold():
...
+
+class DogStream(CatStream):
+    stream = 'dogs'
+    schema = CatStream.schema.copy()
+
+
+DogStream.schema['stream'] = 'dogs'
+
+
+def test_state__doesnt__emit__when__only__one__of__several__streams__is__flushing(capsys):
```
AlexanderMann: nit: underscores 👍

tests/test_target_tools.py
```diff
@@ -70,3 +76,174 @@ def test_loading__invalid__records__threshold():
...
+
+    target_tools.stream_to_target(test_stream(), target, config=config)
+
+    # The final state message should have been outputted after the last records were loaded
+    output = filtered_output(capsys)
+    assert len(output) == 1
+    assert json.loads(output[0])['test'] == 'state-4'
+
+
+def test_state__emits__most__recent__state__when__final__flush__occurs(capsys):
```
AlexanderMann: nit: underscores 👍

tests/test_target_tools.py
```diff
@@ -70,3 +76,174 @@ def test_loading__invalid__records__threshold():
...
+        json.dumps({'type': 'STATE', 'value': { 'test': 'state-1' }}),
+        json.dumps({'type': 'STATE', 'value': { 'test': 'state-2' }})]
+
+    target_tools.stream_to_target(stream, Target(), {'state_support': False})
+    output = filtered_output(capsys)
+
+    assert len(output) == 0
+
+
+def test_state__emits__only__messages__when__all__records__before__have__been__flushed(capsys):
```
AlexanderMann: nit: underscores 👍

target_postgres/state_tracker.py
```diff
@@ -0,0 +1,78 @@
+from collections import deque
+import json
+import sys
```
AlexanderMann: nit: alphabetize the imports in here 👍

target_postgres/state_tracker.py
```diff
@@ -0,0 +1,78 @@
...
+            if force or stream_buffer.buffer_full:
+                self.target.write_batch(stream_buffer)
+                stream_buffer.flush_buffer()
+                self.stream_flush_watermarks[stream] = self.stream_add_watermarks[stream]
+
+        self._emit_safe_queued_states(force=force)
+
+    def handle_state_message(self, value):
+        if self.emit_states:
+            self.state_queue.append((value, self.message_counter))
```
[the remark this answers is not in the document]
airhorns: I agree, but I did it below in the `_emit_safe_queued_states` function so that the diff is done as late as possible. That way it's comparing states it would actually emit and skipping any intermediates in the queue that might be different but not matter.

tests/test_target_tools.py
```diff
@@ -70,3 +76,174 @@ def test_loading__invalid__records__threshold():
...
+
+    target_tools.stream_to_target(stream, Target())
+    output = filtered_output(capsys)
+
+    assert len(output) == 2
+    assert json.loads(output[0])['test'] == 'state-1'
+    assert json.loads(output[1])['test'] == 'state-2'
+
+
+def test_state__capture__can__be__disabled(capsys):
```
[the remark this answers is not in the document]
airhorns: Ah that makes sense, I didn't get that off the bat. I will update!

target_postgres/state_tracker.py
```diff
@@ -0,0 +1,78 @@
+from collections import deque
+import json
+import sys
+import singer.statediff as statediff
+
+
+class TargetError(Exception):
```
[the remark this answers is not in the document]
airhorns: Ok, I added that file, definitely feels cleaner

target_postgres/state_tracker.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def _emit_safe_queued_states(self, force=False):
+        # State messages that occured before the least recently flushed record are safe to emit.
+        # If they occurred after some records that haven't yet been flushed, they aren't safe to emit.
+        # Because records arrive at different rates from different streams, we take the earliest unflushed record as the threshold for what
+        # STATE messages are safe to emit.
+        all_flushed_watermark = min(self.stream_flush_watermarks.values(), default=0)
+        emittable_state = None
+
+        while len(self.state_queue) > 0 and (force or self.state_queue[0][1] <= all_flushed_watermark):
```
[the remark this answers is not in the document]
airhorns: I agree, I think the tuple is unnecessarily opaque! Switched to a dict.

tests/test_target_tools.py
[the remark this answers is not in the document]
airhorns: 💯

airhorns:
Comments addressed :)

AlexanderMann:
Looks good @airhorns! I'm going to leave this open till end of day to allow others to comment if there's dissent. Pending none, I'll merge!

(PS: like the exceptions module 🥇 )

target_postgres/stream_tracker.py
```diff
@@ -0,0 +1,74 @@
...
+import json
+import singer.statediff as statediff
+import sys
+
+from target_postgres.exceptions import TargetError
+
+
+class StreamTracker:
+    """
+    Object to track STATE messages coming in from the tap and which streams need to be flushed before they can be safely emitted.
```
awm33: nit: Should this be updated to reflect the new class? It does more than tracking state messages.

awm33:
@airhorns I just had one quick comment on updating the docs to that class, but otherwise looks great!

This is an amazing contribution!!!

target_postgres/stream_tracker.py
```diff
@@ -0,0 +1,74 @@
...
+        # If they occurred after some records that haven't yet been flushed, they aren't safe to emit.
+        # Because records arrive at different rates from different streams, we take the earliest unflushed record as the threshold for what
+        # STATE messages are safe to emit.
+        all_flushed_watermark = min(self.stream_flush_watermarks.values(), default=0)
+        emittable_state = None
+
+        while len(self.state_queue) > 0 and (force or self.state_queue[0]['watermark'] <= all_flushed_watermark):
+            emittable_state = self.state_queue.popleft()['state']
+
+        if emittable_state:
```
AlexanderMann: @airhorns just noticed this but shouldn't this emit _all_ states which are safe to emit? ie, this conditional block should really be wrapped up underneath the `while` loop?

AlexanderMann: I'm thinking that the bug here would be if I had a couple states which became safe to emit, this would only emit the latest each time. Not sure if that's expected?

airhorns: My reading of the Singer spec was that this behaviour is ok (and in some circumstances desirable), but, I am totally open to changing this. The reason I think it's ok is that because the state messages are opaque, downstream consumers of them can't really presume to do any merging or munging of the states, so the tap itself has to emit a full state each time. I also think it's desirable to not emit superfluous states because they probably trigger some sort of persistence (it is state after all) downstream, so minimizing the amount of persistence that happens seems ideal. In the instance where a tap emits one state per API call, this is potentially a huge reduction in STATE messages emitted and thus STATE messages processed. 

FWIW this matches `target-snowflake`'s `StateBuffer` doodad here: https://gitlab.com/meltano/target-snowflake/blob/master/target_snowflake/target_snowflake.py#L134-136

airhorns:
I added one small commit but big bugfix -- this was only emitting the `value` key of the `STATE` message instead of the whole `STATE` message including the `{"type": "STATE" ...` preamble or any other keys the tap might have added. Oops. I noticed this by actually trying to swap this in for my (rather infantile) infrastructure and figured that out. Updated the tests to expect the whole `STATE` message as well!

airhorns:
Thanks folks!

Issue #102, opened 2019-03-15
Canonicalization of Stream Name (ie, allow uppercase, numeric, invalid string names)


AlexanderMann:
# Motivation

Stream names are not required per the [Singer Spec](https://github.com/singer-io/getting-started/blob/master/docs/SPEC.md) to conform to anything _other_ than to be `"string"s`...

Unfortunately for targets, this means that support for various taps requires supporting many varying formats etc.

Hacks and workarounds exist for this, but if a `tap` has a stream name which is upper case, often the user _cannot_ edit/update this without forking the repo etc.

***For those interested, please vote 👍 on this description for interest in this feature.***

## What we currently do

Presently, _any_ string is allowed for _any_ subtable. This means that if we detect a "nested table" in the data, we will create a "sub-table" where we canonicalize the "path" to that particular field.

ie, if we have a stream `foobar` which has an array field called `B@z`, that means the _path_ to the array field is `("foobar", "B@z")`. This path will be canonicalized as `foobar__b_z`.

[The notable exception for this, is that _stream names_ themselves, ie the "root" table name, _must_ be strings which are](https://github.com/datamill-co/target-postgres/pull/60/files#diff-52720fbc0113bc218f3f4719ab96bbc9R57):
- lowercased
- start with a letter
- contain _only_ alphanumeric characters _and_ `_` underscores

## Why we do what we currently do

This logic was introduced in #60. This PR sought to allow for sub tables to be _reasonable_ names derived from the data/field names provided.

At the time, this was because it was viewed that the _data_ which comes out of a `tap` is well out of the user's control, but that often the `stream` names are already of this format (ie, many API based taps stream's are lowercase etc.).

An attempt was made to make it such that stream names could be of any string format. However it was more work to do that at the time.

## Proposals

Table name mappings are currently stored on the "root" table in a db. ie, the `stream` name table. These mappings take a `path` to a `canonicalized name`.

ie, `(a, b, c, d)` -> `a__b__c__d` etc.

### Dedicated Metadata Table

We could move this mapping out of the JSON blob stored in the table's comment and into a dedicated table. Something akin to `_datamill_table_metadata`.

### Load Table Mappings

Each table's metadata could store the `path`. When `target_postgres` starts, it can list all tables present in the remote, iterate through them and fetch the `raw_name` thus building up a cache in memory of our `path` to `canonicalized name`.

This could even be done via sql with something like:

```sql
SELECT CAST(obj_description(oid) AS json)->'path'
[...]

PR #131, opened 2019-07-30
Feature/stream name  root table  canonicalization

AlexanderMann:
# Motivation

https://github.com/datamill-co/target-postgres/issues/102

- Provide logic which works in `target-redshift` as well as `target-postgres`
  - works across all versions
- Migrates existing remote databases seamlessly
- Existing tests remain untouched

## Notes

This will break instances where _multiple_ versions of `target-postgres` are trying to upsert _different_ streams (taps) to the same remote.

I'm unsure as to whether we _want_ to support this use case.

ie, as a user, I have a long running tap persisting data to postgres db A, via `target-postgres v0.1.8`. I also want to be able to run the latest `target-postgres v0.1.X` to persist another `tap` to db A.

## Suggested Musical Pairing

https://soundcloud.com/snottynoserezkids/sets/snotty-nose-rez-kids-lp

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,6 +229,7 @@ jobs:
           name: Data -> Target
           command: |
             source venv--target-postgres/bin/activate
+            pip install .
             cd /code/.circleci/integration
 
             cat /code/artifacts/data/tap | target-postgres --config target-config.json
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -255,6 +256,7 @@ jobs:
           name: Repeatability of Data -> Target
           command: |
             source venv--target-postgres/bin/activate
+            pip install .
```
AlexanderMann: unsure as to _why_ this is needed but somehow the `virtualenv` ended up getting borked somewhere along the way.

AlexanderMann:
Of note: I realized recently that we assume _all the time_ that there is only **one** target running per db all the time. This is due to us using transactions on the db, so something _should_ just fail quickly if there's a conflict etc.

PR #136, opened 2019-08-07
Add an initial_sql config option to execute arbitrary SQL upon boot

airhorns:
This lets users execute some SQL before target-postgres spins up and does it's thing. This is useful for SQL like `SET ROLE` or setting up encoding business, and I think will end up being generally useful for other stuff I haven't (and don't need to) predict.

For me, this was prompted by needing to `SET ROLE` on Google Cloud SQL. On some postgres deployments, target-postgres may be running as a user with limited permissions, or as a user different than the downstream consumer of the data. In these instances it's important that the table owner of the target-postgres created tables is controllable, especially in deployments where the superuser isn't in the developers control, like Google Cloud SQL. `SET ROLE` is the only way to really control the table owner of the created tables so that other users may be able to grant permissions on those tables outside of the normal `target-postgres` lifecycle. See https://stackoverflow.com/questions/2276644/list-all-tables-in-postgresql-information-schema for a more thourough description of the pieces at play here.


awm33:
@airhorns I could see running SQL before or after a target run as useful, though I'm a little concerned about this loading to scope bloat. I've typically handled this sort of stuff after the target run using something like [dbt](https://www.getdbt.com/). I've also seen people use the SQL Airflow operators to run similar commands.

airhorns:
@awm33 In principle I totally agree -- I don't think `target-postgres` is the right place to do transforms or other work that comes after the E in ETL and, and I'm actually using dbt and airflow already for those downstream doodads! 

The issue is the `SET ROLE` thing. If those downstream SQL steps aren't running as postgres superusers (which on Cloud SQL and RDS is mandatory) they can't change the table ownership or permissions of the tables created by `target-postgres`, which makes it well, hard for them to do any of the useful-but-out-of-scope stuff. Have you folks run into this issue?

awm33:
@airhorns We haven't. Why are you running set role during singer load? We usually set roles and permissions as dbt hooks. We also usually run singer and dbt as the same user. A script in-between those steps could also help manage your roles, though this mostly just seems like you need to look at how your handling permissions. We typically have a "raw zone" that only singer, dbt, and select users have access to. We then transform those tables into another schema(s) using dbt. Sometimes that's just a `select * from table` in order to move it over and apply permissions.

airhorns:
> We also usually run singer and dbt as the same user. 

This is the difference -- we're running singer and dbt as different users so we can workload manage them differently and use other tooling to identify which tools are generating which expensive queries, and being able to instantly see which tool is running what query by having them use different users has been useful for me in the past. This is why I need to `SET ROLE` for the singer user to a shared role, so that it can run as it's own user but share ownership of the created tables with the other users that can administer that raw zone as you call it. 

I am 100% onboard with using tools outside of `target-postgres` to manage permissions, but, that's only possible with some connection setup ahead of time from what I understand. I'm ok if this is a wontfix for me, I'll just be forced to run it as the same user or fork which is regrettable but understandable.

awm33:
@airhorns Okay, that makes a bit more sense now and you've made the case. I just request a slight change/addition. I think  if we use`before_run` and `after_run` might be more extensible. Then if we ever decide to add more SQL hooks, there's a slight convention. Eg `before_batch_write` and `after_batch_write`

airhorns:
🤘 thanks for bearing with me @awm33 ! I've updated this to use that nomenclature and added `after_run_sql` as well.

awm33:
@airhorns Looks good, thank for the contribution! :)

Issue #141, opened 2019-08-22
Confusion around STATE messages


robmoore:
We have been attempting to use the new `STATE` message support in `target-postgres` to enable `LOG_BASED` syncing. We use `tap-postgres` as the corresponding tap and `meltano` to orchestrate the ELT process. 

What we are experiencing is captured in a [meltano](https://gitlab.com/meltano/meltano/issues/884) ticket. However, based on input from a developer on the project, it appears that the root issue is due to a difference in how this target is handling STATE versus how it is handled elsewhere. Quoting from that ticket:

> Looking at https://github.com/datamill-co/target-postgres/pull/130 it seems like they are indeed  sending the full STATE message downstream, whereas Singer themselves only write the value  out in their targets.
> See https://github.com/singer-io/target-csv/blob/a98d4633a97e3b7183a99a4f790722535b97c238/target_csv.py#L91 for target-csv example.

I'd also note that the [`singer-target-template`](https://github.com/singer-io/singer-target-template/blob/master/%7B%7Bcookiecutter.project_name%7D%7D/%7B%7Bcookiecutter.package_name%7D%7D/__init__.py#L77) and [`target-gsheet`](https://github.com/singer-io/target-gsheet/blob/master/target_gsheet.py#L168) duplicate the approach taken by `target-csv`.

awm33:
@robmoore Though the feature was added with some guidance by us, we don't use this feature since we record state at the end from a parent process, though would like to switch to using it for better partial recovery from a target failure.

I think a simple fix would be adding an output format option to the configuration.

@airhorns May have some thoughts too.

Some background on the why Data Mill don't use targets to track state (though may change):

One of the issues with STATE management is that beyond the message, there hasn't been an official spec on using it, just some examples, most not using a complex target. Stitch also doesn't really use Singer targets, it's "target" dumps the schema and records to a queue and a fairly different proprietary tech stack loads data into destinations. Not much has been thought of is using a STATE during the actual load. We also will load against different targets and track messages rather than STATE in that case. Eg with only state tracking through a target, you can't load into two different sources (like S3/Athena and Redshift using only one tap run).

airhorns:
[...]

PR #142, opened 2019-08-23
Emit STATE values to stdout instead of the whole state message

airhorns:
This fixes https://github.com/datamill-co/target-postgres/issues/141 and should improve https://gitlab.com/meltano/meltano/issues/884.

Before this change, target-postgres worked slightly differently than the other taps out there and wrote incoming STATE messages to stdout as the full content of the mesage, including the `{"value": ...}` wrapper around the opaque STATE JSON. Other taps, including the official `singer-target-template`, don't write out that `value` wrapper, and just write the JSON blob contained in it.

This brings target-postgres inline with the other targets and the defacto rule.

### DIFF
--- a/target_postgres/stream_tracker.py
+++ b/target_postgres/stream_tracker.py
@@ -50,9 +50,9 @@ class StreamTracker:
 
         self._emit_safe_queued_states(force=force)
 
-    def handle_state_message(self, value):
+    def handle_state_message(self, line_data):
         if self.emit_states:
-            self.state_queue.append({'state': value, 'watermark': self.message_counter})
+            self.state_queue.append({'state': line_data['value'], 'watermark': self.message_counter})
             self._emit_safe_queued_states()
 
     def handle_record_message(self, stream, line_data):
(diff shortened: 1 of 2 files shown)


airhorns:
Also of note is that this is a breaking change and I think merits a new version. Sorry about that!

aroder:
Sounds like it would break existing versions so major bump https://semver.org/.

Unless you put it behind a flag, then minor bump. 

Then in the next major release you could remove the flag and make it the default behavior

awm33:
@AlexanderMann @airhorns @aroder I don't really have a strong release opinion, but it seems the options are:
1) Just merging it and doing minor release, it seems like only @airhorns was using it anyway?
2) Merge and make it conditional to a config options, doing a minor release
3) Major release and sticking true to the letter of semver

I just grab the latest version by default for most of my ad-hoc work, then lock in for client prod pipelines, so it all works fine for our work.

airhorns:
My guess is that the major version bump would cause more pain for users who have to dive in and figure out what the breaking change was, which in most cases would be for a brand new feature they didn't use before, so I'd say a minor change would be the pragmatic approach.

AlexanderMann:
I'm voting for putting this into a new minor release as this isn't _really_ a patch, and I think anyone who is leveraging this functionality...is already finding it "broken". I'll wait till end of week to release in case others have strong opinions, but in my mind this would end up in `v0.2.0`

PR #149, opened 2019-09-10
Emit STATE messages if only one of many registered streams is adding records

airhorns:
This fixes a small but annoying bug in the STATE message support for taps that output records to many streams at once, and especially taps that emit a whole whack of schemas at the start of the run, and then emit all the records for each stream one by one.

Before this change, the safe-to-emit-STATE-message watermark was the global minimum of each stream's most recently flushed watermark. This works well when you have many streams emitting records at once due to the whole "don't know which streams each STATE message covers so we have to be sure all records from before the STATE message was emitted have been flushed before passing along the STATE message" thing. The add watermark rises higher than the flushed watermark as records are buffered, and then the flush watermark catches up when a flush happens, moving the global minimum up and permitting new STATE messages to be emitted.

Before this change, upon receiving a SCHEMA message for a new stream, the stream is registered, and a most recently added and most recently flushed watermark of 0 is set for that stream. That makes sense because no records for that stream have been added yet, so no records have been flushed yet, so the flush watermark is set to the same thing that the add watermark is. All the SCHEMA message sets up is an expectation for what records might (or might not ever) come. For the purposes of emitting STATE messages, the SCHEMA message doesn't really hold any value, as STATE is really only relative to records.

So, the issue is that the previous version of this code initialized each stream's most recently flushed watermark to 0 upon receiving the SCHEMA message, which locked the global minimum flush watermark at 0 until any records for that stream arrived. If the tap works in such a way that it moves through each stream making API calls for that stream, this means that until the very last stream is processed, that stream's flush watermark at 0, as no records have been flushed yet, but that also means the global flush watermark is at 0 too. This leads to no state messages being output until the end. D'oh.

This changes things around in the tracker class to not include streams who haven't yet seen records to not be included in the global threshold calculation. This has the effect of allowing straightforward initialization on SCHEMA messages while allowing streams to be processed incrementally and STATE messages to be emitted without including not-yet-in-use streams. The code to do this is pretty verbose because the issues are subtle, so I've tried to comment a bit more than usual and be explicit.

I'm a little concerned about this whole thing as I think this is beginning to violate the KISS rule. I try to abide by the whole "write code only as half as smart as you are because you need to be twice as smart as the code to debug it", and, based on the amount of issues coming out of this, it would appear I am at the limits of my smartness. If anyone has suggestions for a simpler approach or a better testing strategy I am all ears! Also happy to provide more context or examples to make reviewing a bit easier. Of note is that I was able to TDD this fix and write the failing test first, then apply this fix to fix the test unchanged.

### DIFF
--- a/target_postgres/stream_tracker.py
+++ b/target_postgres/stream_tracker.py
@@ -37,23 +37,14 @@ class StreamTracker:
         self.streams[stream] = buffered_stream
         self.stream_flush_watermarks[stream] = 0
 
-    def _flush_stream(self, stream):
-        stream_buffer = self.streams[stream]
-
-        # Flush the buffer if any records have been added to the stream
-        if stream in self.streams_added_to:
-            stream_buffer.flush_buffer()
-            self.stream_flush_watermarks[stream] = self.stream_add_watermarks[stream]
-
     def flush_stream(self, stream):
-        self._flush_stream(stream)
+        self._write_batch_and_update_watermarks(stream)
         self._emit_safe_queued_states()
 
     def flush_streams(self, force=False):
         for (stream, stream_buffer) in self.streams.items():
             if force or stream_buffer.buffer_full:
-                self.target.write_batch(stream_buffer)
-                self._flush_stream(stream)
+                self._write_batch_and_update_watermarks(stream)
 
         self._emit_safe_queued_states(force=force)
 
@@ -70,6 +61,12 @@ class StreamTracker:
         self.streams_added_to.add(stream)
         self.stream_add_watermarks[stream] = self.message_counter
         self.streams[stream].add_record_message(line_data)
+
+    def _write_batch_and_update_watermarks(self, stream):
+        stream_buffer = self.streams[stream]
+        self.target.write_batch(stream_buffer)
+        stream_buffer.flush_buffer()
+        self.stream_flush_watermarks[stream] = self.stream_add_watermarks.get(stream, 0)
 
     def _emit_safe_queued_states(self, force=False):
         # State messages that occured before the least recently flushed record are safe to emit.
--- a/target_postgres/stream_tracker.py
+++ b/target_postgres/stream_tracker.py
@@ -21,22 +21,29 @@ class StreamTracker:
         self.emit_states = emit_states
 
         self.streams = {}
+
+        # dict of {'<stream_name>': number}, where the number is the message counter of the most recently received record for that stream. Will contain a value for all registered streams.
         self.stream_add_watermarks = {}
+
+        # dict of {'<stream_name>': number}, where the number is the message counter of the most recently flushed record for that stream. Will contain a value for all registered streams.
         self.stream_flush_watermarks = {}
 
+        self.streams_added_to = set()  # list of stream names which have seen records
         self.state_queue = deque()  # contains dicts of {'state': <state blob>, 'watermark': number}
         self.message_counter = 0
         self.last_emitted_state = None
 
     def register_stream(self, stream, buffered_stream):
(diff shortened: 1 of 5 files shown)


awm33:
@AlexanderMann Can you please help check the logic here? :) This might be a critical bug for some uses, depending on the tap(s) used.

AlexanderMann:
So my initial concern here was around when a user has the configuration option of `persist_empty_tables` set to `true` and what happens. The concern was around that while we haven't received any records per se from a user, we would have changed the remote db. Presently our batching and flushing logic _shouldn't_ be a concern for this.

AlexanderMann:
Actually, I think that _might_ be viewed as currently working depending on how you look at it? @airhorns @awm33 do we want to consider schema persistence (at least when `persist_empty_tables` is `true`) to _also_ be considered a type of mutation worthy of `STATE`?

target_postgres/stream_tracker.py
```diff
@@ -21,22 +21,29 @@ def __init__(self, target, emit_states):
...
 
+        self.streams_added_to = set()  # list of stream names which have seen records
         self.state_queue = deque()  # contains dicts of {'state': <state blob>, 'watermark': number}
         self.message_counter = 0
         self.last_emitted_state = None
 
     def register_stream(self, stream, buffered_stream):
         self.streams[stream] = buffered_stream
-        self.stream_add_watermarks[stream] = 0
         self.stream_flush_watermarks[stream] = 0
```
AlexanderMann: I think if we actually delay adding the `stream_flush_watermarks` here _as well_, we can avoid entirely the usage of `streams_added_to`.

This would only require need `_flush_stream`'s _original_ logic to be modified as:

```py
self.stream_flush_watermarks[stream] = self.stream_add_watermarks.get(stream, 0)
```

Then `_emit_safe_queued_states`'s _original_ logic wouldn't need to change at all right?

ie, if we only set `stream_flush_watermarks` when something is flushed (which should only typically happen when the `buffer_full` anyways, but I digress) and then `stream_add_watermarks` when something is added we should be in the clear...?

airhorns: I actually had it that way when I first put this patch together where `stream_flush_watermarks` had double meaning and captured both the notion of if a stream had been added to and when that stream had been last flushed. But, I felt that because the code was already pretty complicated (and bug prone) that it would be wise to not layer on a second, only somewhat related meaning to the idea of flushing, and so I added the new set to be extra clear. I think it's a little more verbose and will take a couple extra bytes of memory but is a bit clearer for the next poor soul that comes along to touch this code.

airhorns:
>  do we want to consider schema persistence (at least when persist_empty_tables is true) to also be considered a type of mutation worthy of `STATE`

If I understand correctly I don't think it needs to be considered as such. I just pushed a commit that adds a test to verify that despite these changes the tables are still created when `persist_empty_tables` is true, so we can be confident the functionality isn't broken. I also think that creating those tables shouldn't make STATE get emitted as that complicates the code and I can't figure out a time where it would matter. `target-postgres` is idempotent to the existence of those tables and so whether or not they have been created yet I don't think really matters from a STATE perspective. and we can consider the two pieces of functionality orthogonal. 

I also cleaned up the `_flush_stream` function a little to remove the conditional and make it a bit more descriptive.

airhorns:
Any thoughts on this sucker?

AlexanderMann:
Hey @airhorns, sorry for the delay here. "Real Work" has gotten in the way of open source stuffs (gotta make rent somehow 😉)

I'm planning on devoting next Monday to `target-*` maintenance and will most likely have merges etc. for everything then. There are a number of other housekeeping things which would be good to get in as well.

PR #150, opened 2019-09-10
Remove print statement (that I think was for debugging)

airhorns:
This Is Just To Say, By Harry Brundage

I have consumed
the STDOUT lines
for STATE
in my project

And which
You were probably
Just using
for debugging

Forgive me
But I need them to be JSON
So sweet
And so structured


AlexanderMann:
I'll throw a CI test in place to make sure this doesn't happen in the future...sorry folks 😬

Issue #14, opened 2018-11-01
JSON Schema combinations such as "anyOf" and "allOf" are not supported


awm33:
Allow for [combining schemas](https://json-schema.org/understanding-json-schema/reference/combining.html)

AlexanderMann:
@awm33 `$allOf` seems _more_ or less straightforward. At the very least we should be able to get the _main_ use cases of it narrowed down. ie, things like `string`s with a `maxLength`, etc.

Detecting impossible types is something that might be tricky. It's not something that JSON Schema does by default on it's own, so I can see us being in a situation where we'll need to make error messages better/useful.

For `$anyOf` I think we'll need to understand widening columns or be comfortable doing early days Stitch things like making a separate column for each value based on it's `type` as well.

-------

All in all, if you wanna detail what cases you want to cover here/how you want to proceed I can start taking a stab at this.

AlexanderMann:
## anyOf

When `anyOf` is detected, we:
- perform a column split for _each_ schema found
- persist data based on ***the first schema the provided value matches***
  1. I tell you `foo: {anyOf: [string, integer]}`, I provide you with the value `42`
  2. persist that value to `foo__i`
  3. then I provide you with `'bar'`
  4. persist that value to `foo_s`

## allOf

When `allOf` is detected we simply use the _first_ provided schema and save only that in our `simplify` JSON Schema function.

I can see users _eventually_ wanting to have the picking be smarter than just the first provided (ie, I told you `allOf : string, datetime`, use the `datetime` not the `string` as it's more specific) but I don't think doing anything clever to _start_ here is wise.

awm33:
@AlexanderMann These could also be at the base level, eg a stream shema with:
```
{anyOf:[ {type:object,...}, {type:object,...}, {type:object,...}]}
```
In that case, we may need to only allow for root schema anyOf/allOf/oneOf to be objects. I've seen this for APIs where the schema of an endpoint can vary, like an events endpoint.

awm33:
@AlexanderMann Me using "such as" was lazy of me. There's also "oneOf" and "not", though if "not" is difficult, I'd say not support it.

AlexanderMann:
Guess what?!?!? Now they are 🎉 🎉 🎉

PR #155, opened 2019-10-01
Feature/any of support

AlexanderMann:
# Motivation

#14 

This pr seeks to add _full_ support for `anyOf` to `Target-Postgres`.

`anyOf` clauses are merged/simplified according to the following rules (these _are_ recursive):
- all literals are dedupped
- all objects are merged into the same object schema, with sub-schemas being grouped as simplified `anyOf` schemas
- all iterables' `items` schemas are merged as simplified `anyOf` schemas
- all `anyOf`s are flattened to the topmost
- if there is only a single element in an `anyOf`, that is denested
- if any `anyOf`s are nullable, all are nullable

We also simplify "implicit" `anyOf`s. These are the literal schemas where the `type` field is overloaded with multiple types. ie, something like `"type": ["string", "integer"]`

### DIFF
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -229,19 +229,7 @@ class Cachable(dict):
 
     def _comparator(self):
         if not self._c:
-            item_c = tuple()
-            if is_iterable(self):
-                item_c = self['items']._comparator()
-
-            props_c = tuple()
-            if is_object(self):
-                props_c = tuple(sorted(self.get('properties', {}).items()))
-
-            self._c = (
-                tuple(sorted(get_type(self))),
-                self.get('format', ''),
-                item_c,
-                props_c)
+            self._c = json.dumps(self, sort_keys=True)
 
         return self._c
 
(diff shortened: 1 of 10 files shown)


target_postgres/json_schema.py
```diff
@@ -189,15 +200,43 @@ def make_nullable(schema):
...
+        return self._comparator().__hash__()
+
+    def deepcopy(self):
+        s = deepcopy(self)
+        s._c = self._c
+        return s
+
+    def _comparator(self):
+        if not self._c:
+            self._c = json.dumps(self, sort_keys=True)
```
awm33: Any performance hit here?

tests/unit/test_denest.py
```diff
@@ -280,3 +277,118 @@ def test__records__nested__child_table__a_b_c_e():
 
         assert 'boolean' == record[('g',)][0]
         assert bool == type(record[('g',)][1])
+
+
+def test__anyOf__schema__stitch_date_times():
+    denested = error_check_denest(
+        {'properties': {
```
awm33: 👍 This is our primary test case

target_postgres/json_schema.py
AlexanderMann: Yeah, but just once and only on schema messages (at least should be). The reason for having this is so that this doesn't get run all the time.

PR #156, opened 2019-10-02
Automatically source the virtual env when entering the development docker container

airhorns:
This is a tiny quality of life thing. I think it'd be handy to do this so you don't have to remember to do it yourself! I also updated the readme to use `docker-compose` to exec into the development container so it's also easier to remember and less sensitive to the container naming issue.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -6,6 +6,7 @@ source /code/venv/target-postgres/bin/activate
 
 pip install -e .[tests]
 
+echo "source /code/venv/target-postgres/bin/activate" >> ~/.bashrc
 echo -e "\n\nINFO: Dev environment ready."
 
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -6,6 +6,7 @@ source /code/venv/target-postgres/bin/activate
 
 pip install -e .[tests]
 
+echo "source /code/venv/target-postgres/bin/activate" >> ~/.bashrc
```
AlexanderMann: Brilliant 👍

AlexanderMann:
@airhorns my only concern here _was_ when you need to do something not inside the `venv`, but then I realized that that is far less of an occurrence than using the `venv`, and you can `deactivate`.

PR #169, opened 2020-01-16
Fix/buffer size calculations

AlexanderMann:
# Motivation

There were a string of issues surrounding users having really slow load times for batches which were being cut at _far_ below the "expected" threshold based on the number of _records_ folks presumed to be in a given batch.

The issue ended up being that our `pysize` implementation was ballooning the size of nested records. A more standard approach for setting the buffer size can be found in [Singer's `target-stitch` implementation](https://github.com/singer-io/target-stitch/blob/master/target_stitch/__init__.py#L599). ie, use the non-json-decoded raw line's length. Python stores strings as ~37 bytes + 1 byte per character. Given a json-encoded string will have _massively_ more formatting than a CSV line corresponding, we should be fine to simply use this as a drop in replacement.


target_postgres/singer_stream.py
```diff
@@ -19,6 +19,12 @@
 SINGER_LEVEL = '_sdc_level_{}_id'
 SINGER_VALUE = '_sdc_value'
 
+DATAMILL_RAW_LINE_SIZE = '_dm_raw_line_size'
```
awm33: This almost makes it seem like it's persisted, which it isn't? Maybe just `__raw_line_size`? As long as it's not persisted, I wouldn't worry about namespacing.

target_postgres/target_tools.py
```diff
@@ -91,6 +91,7 @@ def _line_handler(state_tracker, target, invalid_records_detect, invalid_records
                   max_batch_size, line):
     try:
         line_data = json.loads(line)
+        line_data[DATAMILL_RAW_LINE_SIZE] = len(line)
```
awm33: Should we only add this for record types?

AlexanderMann: Makes sense. Easy change to make.

target_postgres/singer_stream.py
AlexanderMann: Good point. `__` should be fine for ensuring the payload itself occludes any conflicting keys.

AlexanderMann:
@awm33 back to you

Issue #178, opened 2020-03-20
Column fails if its value is 1 or 2


davicorreiajr:
I'm have a weird issue: whenever the column value is 1 or 2, I get the following error:
```
target_postgres.exceptions.SingerStreamError: ('Invalid records detected above threshold: 0. See `.args` for details.', [(<ValidationError: '2.0 is not valid under any of the given schemas'
```
It's weird that I get the error for 1 or 2, but from 3 on it seems to work.

If it helps, the generated schema for this column is:
```
{
   "anyOf":[
      {
         "type":"null"
      },
      {
         "type":"number",
         "multipleOf":1e-15
      },
      {
         "type":"string"
      }
   ]
}
```
I'm using `target-postgres` along with `tap-google-sheets`.

AlexanderMann:
@davicorreiajr that is bizarre, but I _imagine_ that has something to do with floating point error/binary number gaps? Our code _doesn't_ actually do those checks, that's Python's implementation of JSONSchema (we try to do as _little_ data munging when validating the schema against the records we ingest).

This looks very familiar: https://github.com/Julian/jsonschema/issues/247

djdevin:
Another postgres target with the same issue, possible solution:

https://github.com/statsbotco/target-postgres/pull/11/files

AlexanderMann:
@djdevin ah, interesting. I _think_ the best way to solve this, with the smallest footprint is prolly just to use the suggestion straight from the Python JSON docs:

```py
>>> import decimal
>>> json.loads('1.1', parse_float=decimal.Decimal)
Decimal('1.1')
```

With that sort of change, handling this issue _might_ be trivial.

AlexanderMann:
If someone wants to put up a PR for the above which adds a test for the original problem AND uses the simple `json.loads` trick, I'm happy to review and try to get this out early next week. If not, I'll try and get the change in myself! 😄

davicorreiajr:
Nice, thanks guys!

I'll try and do it.

AlexanderMann:
This just needs to be deployed now. Will close once deployed.

PR #179, opened 2020-03-28
Fix parse to properly support multipleOf (to avoid validation error)

davicorreiajr:
# Motivation

#178

### DIFF
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -1,6 +1,7 @@
 from copy import deepcopy
 import json
 import re
+import decimal
 
 from jsonschema import Draft4Validator
 from jsonschema.exceptions import SchemaError
@@ -20,7 +21,8 @@ _PYTHON_TYPE_TO_JSON_SCHEMA = {
     float: NUMBER,
     bool: BOOLEAN,
     str: STRING,
-    type(None): NULL
+    type(None): NULL,
+    decimal.Decimal: NUMBER
 }
 
 
--- a/target_postgres/json_schema.py
+++ b/target_postgres/json_schema.py
@@ -1,7 +1,7 @@
 from copy import deepcopy
+import decimal
 import json
 import re
-import decimal
 
 from jsonschema import Draft4Validator
 from jsonschema.exceptions import SchemaError
--- a/tests/unit/test_BufferedSingerStream.py
+++ b/tests/unit/test_BufferedSingerStream.py
@@ -140,7 +140,6 @@ def test_add_record_message__multipleOf_invalid_record():
 
     assert singer_stream.peek_invalid_records()
     assert singer_stream.count == 0
-    assert [] == missing_sdc_properties(singer_stream)
 
 
 SIMPLE_ALLOF_SCHEMA = {
--- a/tests/unit/test_BufferedSingerStream.py
+++ b/tests/unit/test_BufferedSingerStream.py
@@ -1,9 +1,10 @@
+from decimal import Decimal
 from copy import deepcopy
 
 import pytest
 
 from target_postgres import singer
-from target_postgres.singer_stream import BufferedSingerStream, SingerStreamError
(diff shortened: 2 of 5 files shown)


tests/unit/test_BufferedSingerStream.py
```diff
@@ -75,6 +76,73 @@ def test_add_record_message__invalid_record():
...
+                    'stream': stream_name,
+                    'record': {'multipleOfKey': value},
+                    'sequence': 0,
+                    RAW_LINE_SIZE: 100
+                }
+            )
+
+    assert singer_stream.peek_invalid_records()
+    assert singer_stream.count == 0
+    assert [] == missing_sdc_properties(singer_stream)
```
AlexanderMann: I think we can drop this. It's not really adding value to the test overall.

tests/unit/test_BufferedSingerStream.py
```diff
@@ -75,6 +76,73 @@ def test_add_record_message__invalid_record():
...
+    multiple_of_values = ['1', '2', '3', '4', '5', '1.1', '2.3', '1.23456789', '20', '100.1']
+
+    for value in multiple_of_values:
+        singer_stream.add_record_message(
+            {
+                'type': 'RECORD',
+                'stream': stream_name,
+                'record': {'multipleOfKey': Decimal(value)},
+                'sequence': 0,
+                RAW_LINE_SIZE: 100
```
AlexanderMann: I think we can drop this. It's not really adding value to the test overall.

tests/unit/test_json_schema.py
```diff
@@ -31,6 +32,8 @@ def test_python_type():
            == json_schema.STRING
     assert json_schema.python_type('world') \
            == json_schema.STRING
+    assert json_schema.python_type(decimal.Decimal(1)) \
+           == json_schema.NUMBER
```
AlexanderMann: Excellent

target_postgres/json_schema.py
```diff
@@ -1,6 +1,7 @@
 from copy import deepcopy
 import json
 import re
+import decimal
```
AlexanderMann: This isn't documented anywhere, so I wouldn't expect you to know this one, but we try to alphabetize our `import`s. The breakdown is:

- Python
- 3rd Party
- Our code

tests/unit/test_BufferedSingerStream.py
davicorreiajr: actually, we need to keep this. Because:

1. you add it here: https://github.com/datamill-co/target-postgres/blob/master/target_postgres/target_tools.py#L139
2. and then, when you call `get_line_size` here https://github.com/datamill-co/target-postgres/blob/master/target_postgres/singer_stream.py#L152
3. we return `line_data.get(RAW_LINE_SIZE)` (doing `json.dumps(line_data)` crashes, because of the Decimal)

AlexanderMann: Ah, damnit, sorry, fat fingers. I meant to mark the _similar line_ as the other comment I left. You're absolutely _right_ that _this_ line is necessary.

tests/unit/test_BufferedSingerStream.py
```diff
@@ -75,6 +76,72 @@ def test_add_record_message__invalid_record():
...
+            }
+        )
+
+    assert not singer_stream.peek_invalid_records()
+    assert singer_stream.count == len(multiple_of_values)
+    assert [] == missing_sdc_properties(singer_stream)
```
AlexanderMann: @davicorreiajr it's this line, we can remove it same as the other, sorry for the confusion and appreciate your patience!

davicorreiajr: no worries, fixing now

AlexanderMann:
Will merge and hopefully deploy later this week!

PR #190, opened 2020-08-05
Feature: Use Postgres env vars in absence of explicit config

crwr45:
This allows the tap to be added into existing Postgres environments
more easily, and reduces configuration duplication.

The pre-existing config still takes precedence, and the defaults
in the absence of config and env vars are what they were before.

Signed-off-by: crwr45 <charlie.wheelerrobinson@gmail.com>


target_postgres/__init__.py
```diff
@@ -40,6 +38,6 @@ def main(config, input_stream=None):
 
 
 def cli():
-    args = utils.parse_args(REQUIRED_CONFIG_KEYS)
```
AlexanderMann, line 43: So I think we're going to want something akin to this as the failure mode for folks trying to run this when they _don't_ have the proper values present. Maybe a simple helper function which mimics the functionality of `parse_args` but does so with the fallbacks of the `environ`?

AlexanderMann:
@awm33 any movement here, anything that should be done?

laurentS:
I think this would also help with security and following the 12 factor principles, as it removes the need for writing a password to a file

ericboucher:
@crwr45 could you have a quick look and make sure the tests are passing? Thanks!

Issue #185, opened 2020-06-16
Missing pg_config causes installation failure when using pip


ns-iknox:
Issue description:
Installation fails for `pip install singer-target-postgres` when installing the dependency `psycopg2` in an environment that doesn't have `pg_config` in the path.

Proposed solution:
Conditionally install the binary distribution (`psycopg2-binary`) if `pg_config` isn't found in the path.

ns-iknox:
After some more research it's clear that regardless of installation issues, we should only be using the sdist.

From the [psycopg2 docs:](https://www.psycopg.org/docs/install.html#binary-packages)

> **Note** The `psycopg2-binary` package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements.
If you are the maintainer of a published package depending on psycopg2 **you shouldn’t use `psycopg2-binary` as a module dependency**. For production use you are advised to use the source distribution.

ns-iknox:
Closing Issue

INGCRENGIFO:
> After some more research it's clear that regardless of installation issues, we should only be using the sdist.
> 
> From the [psycopg2 docs:](https://www.psycopg.org/docs/install.html#binary-packages)
> 
> > **Note** The `psycopg2-binary` package is meant for beginners to start playing with Python and PostgreSQL without the need to meet the build requirements.
> > If you are the maintainer of a published package depending on psycopg2 **you shouldn’t use `psycopg2-binary` as a module dependency**. For production use you are advised to use the source distribution.

Hello, how did you solve it?

HerbiePorter:
@INGCRENGIFO You need to ensure you have the postgres client installed.  

If you're using a Mac you can use Brew: `brew install postgres`

geofflangenderfer:
on ubuntu, `sudo apt install libpq-dev`

vadim82:
How do we get this working on windows? I am getting this error when I try to pip install this package in windows.
```
Error: pg_config executable not found.
```

PR #195, opened 2020-12-31
add libpq dependency

geofflangenderfer:
#185

I'm unable to run without installing this dependency


AlexanderMann (commented):
This is a great addition, we could probably link folks to the PsycoPG2 docs as well, but this seems worthwhile to get in to start. We've seen a couple people having trouble with this recently.

AlexanderMann (approved):

PR #200, opened 2021-03-12
Add SQL files for before/after hooks (in addition to SQL strings)

dkarzon:
Added the ability to use external files as before/after hooks instead of needing to put sql in the json configs because we currently use the after hook to do some basic data transformation once it gets into the database such as additional calculated columns (ie. Google Analytics column for a clean url on pageviews) and this SQL gets quite large when it needs to be on 1 line in a json file.

Example config file using this:
   ```json
   {
     "postgres_host": "localhost",
     "postgres_port": 5432,
     "postgres_database": "my_analytics",
     "postgres_username": "myuser",
     "postgres_password": "1234",
     "postgres_schema": "mytapname",
     "after_run_sql_file": "target_postgres_after.sql"
   }
   ```

Example SQL file:
```sql
ALTER TABLE ga_pageviews ADD COLUMN IF NOT EXISTS ga_cleanpage TEXT;
UPDATE ga_pageviews SET ga_cleanpage = CASE WHEN strpos(ga_pagepath, '?') = 0 THEN ga_pagepath ELSE left(ga_pagepath, strpos(ga_pagepath, '?') - 1) END WHERE ga_cleanpage is NULL;
```


laurentS:
This looks like a fairly clean solution. LGTM :)

target_postgres/target_tools.py
```diff
@@ -188,3 +188,10 @@ def _run_sql_hook(hook_name, config, target):
         with target.conn.cursor() as cur:
             cur.execute(config[hook_name])
             LOGGER.debug('{} SQL executed'.format(hook_name))
+
+    hook_file = hook_name + '_file'
```
ericboucher, line 192: Should we add a warning or raise an error if both a string and a file are provided? It might create conflicts and we don't have a way to say if one should take precedence or be run first.

@laurentS wdyt?

dkarzon: The text property version in config will always run before the file version so I don't think that will be much of an issue with regard to the ordering.

ericboucher:
@dkarzon sorry for taking over your PR, I'm trying to get the tests to run correctly on PRs coming from forks. I'll cleanup everything once I figure it out, promise! :)

ericboucher (approved):

PR #201, opened 2021-03-12
Setup a vscode devcontainer for easier development

dkarzon:
This let's you run your VS Code dev environment inside a docker container so that you don't need to install all of the dependencies. This will especially help new contributors getting setup on the project.


AlexanderMann:
Thanks @dkarzon! Really appreciate the contribution. I'm not personally familiar with `devcontainer`s over `docker-compose`. Any links/reading you'd recommend?

Additionally, for a change of this type, could you add in changes to the README to alert folks as to how to use this?

dkarzon:
Basically this lets developers run an instance of VS Code inside a docker container for this project with the correct environment preconfigured so there is no need to worry about versioning conflicts etc.

For me this helped me be able to build this on a windows machine without having to install python and all of the dependencies natively.

More info here: https://code.visualstudio.com/docs/remote/containers

Issue #194, opened 2020-11-11
Erro with the target when paired with zendesk tap


j-oflaherty:
Hello. I've run into an issue when trying to use this target alongside the zendesk tap, when the satisfaction_ratings stream provided by the latter is selected. Two errors appear to be happening at the same time, but only one happens within this target's files.

```
INFO Querying for satisfaction ratings between 2020-10-19T23:59:59Z and 2020-11-11T19:13:02Z
CRITICAL 'properties'
Traceback (most recent call last):
  File ".virtualenvs/target-postgres/bin/target-postgres", line 8, in <module>
    sys.exit(cli())
  File "~/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/__init__.py", line 45, in cli
    main(args.config)
  File " ~/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/__init__.py", line 39, in main
    target_tools.main(postgres_target)
  File "(directory)/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/target_tools.py", line 28, in main
    stream_to_target(input_stream, target, config=config)
  File "(directory)/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/target_tools.py", line 77, in stream_to_target
    raise e
  File "(directory)/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/target_tools.py", line 58, in stream_to_target
    _line_handler(state_tracker,
  File "(directory)/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/target_tools.py", line 123, in _line_handler
    buffered_stream = BufferedSingerStream(stream,
  File "(directory)/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/singer_stream.py", line 47, in __init__
    self.update_schema(schema, key_properties)
  File "(directory)/singer_testing/.virtualenvs/target-postgres/lib/python3.8/site-packages/target_postgres/singer_stream.py", line 75, in update_schema
    properties = self.schema['properties']
KeyError: 'properties'
```
This is the first error that occurs. After a second error with the tap, this is what it's displayed:
```
BrokenPipeError: [Errno 32] Broken pipe
Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
[...]

PR #202, opened 2021-03-18
Zendesk-Tap Bug fixed

j-oflaherty:
Hi,

Here is the solution I found on issue #194. I don't know if there's a better one to solve this issue, but so far I haven't run into other problem with it.


leo-schick:
@j-oflaherty I do not see this here as a good solution. You just select the first entry of the `"anyOf"` array, but actually, you should iterate through the array to determine the right schema. It might happen that the array elements are switched and then it will fail again.

Issue #203, opened 2021-04-04
Performance issues loading 5M rows from CSV file


laurentS:
I'm using `target-postgres` with `meltano` and `tap-spreadsheets-anywhere` to load a simple CSV file containing 2 columns of strings (each has fixed width, all values populated). The operation was initially very slow (over 45 minutes, on a single machine 4CPU/32GB RAM, no networking involved).

When first running the tap to target process, it processed the file in batches of 200k rows, rough metrics:
- tap: 20 secs/batch
- target: ~85s/batch

With a few `time()` calls, I identified that most of the time was spent in `_serialize_table_records`, specifically in this loop: https://github.com/datamill-co/target-postgres/blob/9c095d91e215f932caa897a6587c6dc6278db8cf/target_postgres/sql_base.py#L728

Then throwing `cProfile` at the loop, I got the following results:
![image](https://user-images.githubusercontent.com/1745184/113487429-9b3f5380-94a7-11eb-8916-94ee185e9f3a.png)

My impression was that there is only a limited combination of arguments to `_serialize_table_record_field_name` so that it can be cached to avoid recalculating over and over.
The same seems to apply to `serialize_table_record_datetime_value` (in my tests, each record has the same value in `_sdc_batched_at`).

The last bit of excessive time is spent in `deepcopy` which is a weirdly slow function. Replacing it with pickling/unpickling saves a surprising amount of time.

With these changes, the target runs the same batch in about 18s instead of 85.

I'm filing a PR that addresses these issues.

laurentS:
Fixed in #204

PR #204, opened 2021-04-04
Multiple performance improvements

laurentS:
This PR addresses the changes proposed in #203 

As described in the issue, loading data seemed unnecessarily slow. On a test CSV file containing 50k rows, each with 2 cols of fixed length text, `target-postgres` was spending about 20 seconds loading the data.

Some profiling helped identify the bottlenecks, as described in the issue.

This PR proposes 3 distinct improvements, in decreasing order of time savings (based on my test case). With `cProfile` instrumentation, the loading took approx 32 seconds instead of 20:

- There is only a limited combination of arguments to `_serialize_table_record_field_name` so that it can be cached to avoid recalculating over and over. The method was called for each field of each row, including all 5 metadata fields. In my example, this led to about 350k calls. With `functools.lru_cache` applied, `cache_info` showed that only 7 calls were really made, all remaining calls resulted in cache hits. My test data is extremely consistent, so might be a best case, but my understanding is that this should result in at most `number_of_fields * possible_data_types` actual calls. If multiple batches are required to load the data, each batch will recreate a fresh cache. My fix to allow caching feels a bit hackish, but it seems to work (improvement suggestions welcome!). This saves about 15s/32 (with profiling on).
- The same seems to apply to serialize_table_record_datetime_value (in my tests, each record has the same value in _sdc_batched_at). The call to `arrow` is pretty expensive. In my testing, the exact same value was formatted once per row. Applying caching on this seems straightforward. In my sample, it led to about 14s/32 of time savings.
- The last bit of excessive time is spent in `deepcopy` which is a weirdly slow function. Replacing it with pickling/unpickling saved about 1.5s again.

In the end, with these speedups applied, the same data now loads in under 4s instead of the original 20s.

### DIFF
--- a/target_postgres/postgres.py
+++ b/target_postgres/postgres.py
@@ -1,5 +1,6 @@
 from copy import deepcopy
 import csv
+from functools import lru_cache
 import io
 import json
 import logging
@@ -19,6 +20,14 @@ from target_postgres.sql_base import SEPARATOR, SQLInterface
 
 RESERVED_NULL_DEFAULT = 'NULL'
 
+@lru_cache(maxsize=128)
+def _format_datetime(value):
+    """
+    Format a datetime value. This is only called from the
+    PostgresTarget.serialize_table_record_datetime_value
+    but this non-method version allows caching
+    """
+    return arrow.get(value).format('YYYY-MM-DD HH:mm:ss.SSSSZZ')
 
 def _update_schema_0_to_1(table_metadata, table_schema):
     """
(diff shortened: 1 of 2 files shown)


berenddeboer:
This fix is remarkably effective. It cuts down my processing time in half.

I'm reading 102 smallish files (about 1MB each) from S3, and writing to AWS Aurora Serverless. I'm writing about 200,00 records. This patch cuts down my time from 6 minutes to 3 minutes.

berenddeboer:
And another test: 8.8 million rows. Used to take 86 minutes, now down to 57 (while the machine was overloaded doing other stuff). I think this should really be merged!

laurentS:
FWIW, I've forked this repo and added some further performance improvements and memory savings in https://github.com/oviohub/target-postgres/

PR #210, opened 2021-11-04
Map the sql type integer to the integer json type

dkarzon:
Make sure we can map SQL integer types to JSON integer types


laurentS:
@dkarzon can you explain the use case for this? I just do not understand how both sql integer types are used. Not challenging the need for it, just my own understanding!

dkarzon:
@laurentS My use case for this is probably an odd one. I can't remember the exact scenario but I have some transforms on the data between tap and target, one of which was adding a new field to the schema of type integer. Without this fix I think target-postgres was either rebuilding the table or erroring out because it couldn't match the types.

PR #213, opened 2022-04-12
Fix circleCI (bypass need for SSH key)

ericboucher:
Since target-postgres is open-source, we bypass the need for an SSH key and clone the open-source version of the repo directly.


laurentS (approved):
Looks good to go once the tests are green :)

PR #215, opened 2022-04-17
Add option to set pg application_name

laurentS:
Postgresql has a [connection setting](https://www.postgresql.org/docs/12/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-WHAT) which allows identifying which application is connecting to it, which makes it easier to understand logs and debug, etc...

This PR makes it possible to change this setting via the config file. The PR is completely backwards compatible, so not setting this in config won't change any of the current output.


ericboucher:
Looks great! Can you rebase on the latest master? The format of the config readme doesn't seem up to date?

laurentS:
It seems to be based off the latest `master` af4856b from what I can see. I can't spot any difference in the readme between my branch and `master` (except my extra row in the table, of course). Did I miss something?

PR #220, opened 2022-08-28
Updating singer_stream.py to allow schemas without properties

lucas:
While utilizing tap-zendesk (https://github.com/twilio-labs/twilio-tap-zendesk) there are some cases where the generated schema does not include a `properties` key in the dictionary, causing the `properties = self.schema['properties']` to fail.

Given that there is nothing returned from this function, returning early appears to resolve it.

### DIFF
--- a/target_postgres/singer_stream.py
+++ b/target_postgres/singer_stream.py
@@ -71,7 +71,11 @@ class BufferedSingerStream():
 
         # The validator can handle _many_ more things than our simplified schema, and is, in general handled by third party code
         self.validator = Draft4Validator(schema, format_checker=FormatChecker())
-
+        
+        # Some taps do not provide 'properties' key, which causes the next block to fail
+        if not (self.schema and 'properties' in self.schema):
+            return
+        
         properties = self.schema['properties']
 
         if singer.RECEIVED_AT not in properties:


target_postgres/singer_stream.py
```diff
@@ -71,7 +71,11 @@ def update_schema(self, schema, key_properties):
 
         # The validator can handle _many_ more things than our simplified schema, and is, in general handled by third party code
         self.validator = Draft4Validator(schema, format_checker=FormatChecker())
-
+        
+        # Some taps do not provide 'properties' key, which causes the next block to fail
+        if not (self.schema and 'properties' in self.schema):
+            return
+
```
laurentS, line 78: Wouldn't it be easier to write something like this below?
`properties = self.schema.get('properties', {})
`
among others to ensure that the test on `self.key_properties` is executed? (not sure if that situation makes sense logically though)

laurentS (commented):
What do you think of the approach below instead?

target_postgres/singer_stream.py
lucas: I believe that should work just as well for my case, I'll update it tonight.

Thanks @laurentS

laurentS:
Quick sense-check on this, is the absence of `properties` in the schema allowed by the singer spec? I took a quick look, but couldn't find a clear answer, in part because I don't have the data that causes problem. Basically, I'm wondering if the bug is on this end or in the tap you're using.

laurentS:
@lucas would you be able to merge the main branch into this PR to switch to the new dependency management format (pyproject.toml)?

PR #224, opened 2022-11-03
Upgrade dependency management and versions

laurentS:
This PR does a refresher on the package:
- removes the old `setup.py` and replaces it with a shiny `pyproject.toml`
- switches to `poetry` for dependency management. This allows better build reproducibility, version locking, etc... In my opinion, it would also make sense to replace `twine` for uploading to `pypi` with `poetry`, just to have a single tool for everything. I am proposing `poetry` as it is pretty stable and widely adopted at this point.
- updates CI to test against currently maintained versions of `postgresql`. Version `10` will reach its end of life in a few days, but I'll keep it here for now.


laurentS:
With this, the CI checks are still only run on python 3.7. Given how stable the core postgresql is, and the fact that the target does not use much bleeding edge postgresql at all, I suspect there would be more value in testing over all supported versions of python rather than just 3.7.
Any suggestions about how to do this without ending up with 30+ test jobs?

ericboucher (approved):

ericboucher:
@laurentS what do you think about upgrading pyscopg while we're at it? https://github.com/datamill-co/target-postgres/issues/225

laurentS:
I've just updated `pyproject.toml` to use `psycopg2-binary` instead of `psycopg2`.
@WillDaSilva @akhefifderain you both opened issues related to this as mentioned in the previous comment. Can you chime in on the changes proposed in this PR, in particular check that it works with your respective use cases?

akhefifderain:
Seems good to me ! Thank you @laurentS

PR #238, opened 2025-01-06
[Fix] Poetry build tool

### LINKED ISSUE #13: JSON Schema $ref is not supported

SudarshanVS:


SudarshanVS:
Sorry, wrong repo.

